PR #2, opened 2024-11-08
Dependencies, .gitignore, configuration management and examples

cantordust:
- Added `pyproject.toml` and `.gitignore`.
- Updated `README.md` with some information about the dependencies.

Edit [18 Nov 2024]

- Added the Urban-M4 fork of the Mapillary Python SDK as a dependency. The original version does not support recent versions of Python, and lots of dependencies are outdated.
- Added a basic configuration management module. Please read the section about environment variables in README.md.
- Added a simple example that demonstrates how to interface with the Mapillary API via the SDK.
- The data subset for Amsterdam is saved to a separate file (`amsterdam-all.parquet`), which is much faster to load in the `plot_city.ipynb` notebook.
- [Added a licence, more notebooks and functions for downloading images …](https://github.com/Urban-M4/streetscapes/pull/2/commits/33876e6f39f1f60c6d7c65146aeffefb82e6d889) 
[33876e6](https://github.com/Urban-M4/streetscapes/pull/2/commits/33876e6f39f1f60c6d7c65146aeffefb82e6d889)
…from Mapillary and KartaView.

Edit [20 Nov 2024]
- Added functions to download a single image or a batch of images from `Mapillary` or `KartaView`. The setup is very generic and can be extended easily to include other sources. Batch downloads are parallelised.
	- Added a `SourceMap` enum class, which can be used for filtering datasets, correct URL fetching and so on.
- Added a `LICENCE.md` file with the CC BY 4.0 licence text.
- Updated `README.md` file with some more information about how to install and work with `streetscapes`.
- Added an `environment.yml` file for convenience for users who prefer to use a Conda / Mamba environment.
- Updated notebooks:
	- `notebooks/download_city_images.ipynb`: A notebook demonstrating how to download images from `Mapillary` and `KartaView` based on a data file (csv or parquet).
	- `notebooks/mapillary_example.ipynb`: An example of downloading a single image from Mapillary based on coordinates.
- Fixed issues with the configuration module.
- Added a simple logger (necessary for printing non-interleaved output from multiple threads).
- Deleted `requirements.txt` (superceded by `pyproject.toml`).
- Added a CLI command for downloading images from the terminal.

Edit [21 Nov 2024]
- Added a switch to silence the prompt in convert_csv_to_parquet() and a notebook demonstrating how to use the function with that switch.
- Added `ruff` to the list of dependencies.
- Processed all files with `ruff check --fix --select I`.
- Minor fix for logger configuration


LICENCE.md
ClaireDons, line 1: We should add an apache licence here...

README.md
```diff
@@ -1,26 +1,63 @@
...
 
-### Converting global-streetscapes to parquet
+# Installation
+Clone the repository, enter the root directory and use pip to install the `streetscapes` package:
 
-`convert_csv_to_parquet.py` contains the code for reproducing `streetscapes-data.parquet` dataset. 
-For input you will have to download the original csv files from [hugging face](https://huggingface.co/datasets/NUS-UAL/global-streetscapes/tree/main/data). The code looks for these csv files in `../streetscapes-data`, which you may need to create yourself or change the path to the location of the files. 
-To limit size, the dataset contains the following data:
+```shell
+$> git clone git@github.com:Urban-M4/streetscapes.git
```
ClaireDons, line 8: Add that a virtual environment should be created.

README.md
```diff
@@ -1,26 +1,63 @@
...
-    - `segmentation.csv`
-    - `simplemaps.csv`
+### Dependencies
+There are a lot more dependencies in `pyproject.toml` than strictly necessary to run the examples in this repository. They are necessary for running (at least part of) the code in the original `global-streetscapes` repository, specifically the training pipeline (WIP).
 
-It is possible to combine more CSV files if needed. 
+Streetscapes uses a [custom version](https://github.com/Urban-M4/mapillary-python-sdk) of the [Mapillary Python SDK](https://github.com/mapillary/mapillary-python-sdk) which fixes some dependency issues.
 
-### Analysing the data
+## Environment variables
```
ClaireDons, line 19: I think it would be nice if the notebooks that don't require mapillary can still be run without this throwing an error I think.

README.md
```diff
@@ -1,26 +1,63 @@
...
 
-`plot_city.ipynb` shows a simple of example of subsetting the dataset and plotting the data. 
-`subset_data.ipynb` shows an example of subsetting the data for image download, similar to [this example](https://github.com/ualsg/global-streetscapes/blob/main/code/download_imgs/sample_subset_download.ipynb)
+- `MAPILLARY_CLIENT_ID`: A Mapillary client ID string (can be obtained via your [Mapillary account](https://www.mapillary.com/app/)).
+- `MAPILLARY_TOKEN`: A Mapillary token string used for authentication when querying Mapillary via their API.
+- `STREETSCAPES_DATA_PATH`: A directory containing data from the `global-streetscapes` projects, such as CSV files (cf. below). Defaults to `<repo-root>/local/streetscapes-data`.
+- `STREETSCAPES_OUTPUT_DIR`: A directory for output files. Defaults to `<repo-root>/local/output`.
 
-### Acknowledgements/Citation
+## CLI
```
ClaireDons, line 27: I had a few problems running the CLI initially, I think it might be nice to also leave the option of just running the code, especially as there is currently only one command (although it nice that there is space to add more now).

pyproject.toml
```diff
@@ -0,0 +1,41 @@
...
+    "pandas",
+    "pyarrow",
+    "pyproj",
+    "python-decouple",
+    "rasterio",
+    "seaborn",
+    "timezonefinder",
+    "tqdm",
+]
+authors = []
```
ClaireDons, line 34: Perhaps we can already add ourselves here.

requirements.txt
ClaireDons, line 1: Is this file still needed for installing dependencies?

streetscapes/conf.py
```diff
@@ -0,0 +1,28 @@
...
+    config("STREETSCAPES_DATA_PATH", LOCAL_DIR / "streetscapes-data")
+)
+DATA_DIR.mkdir(exist_ok=True, parents=True)
+
+# Output directory
+# ==================================================
+OUTPUT_DIR = Path(config("STREETSCAPES_OUTPUT_DIR", LOCAL_DIR / "output"))
+OUTPUT_DIR.mkdir(exist_ok=True, parents=True)
+
+# Mapillary configuration
```
ClaireDons, line 25: Separating this out so that it is only needed for the notebooks and code that requires it might be nice.

streetscapes/conf.py
ClaireDons, line 1: I had a problem originally where my data was not somewhere "expected" which led to an error, even though I had set it in the .env file: 

`ValueError: '/Users/clairedonnelly/Documents/Urban-M4/streetscapes-data/simplemaps.csv' is not in the subpath of '/Users/clairedonnelly/Documents/Urban-M4/streetscapes' OR one path is relative and the other is absolute.`

This was solved by changing where my data was located but ideally, the data should be able to be anywhere.

streetscapes/conf.py
```diff
@@ -0,0 +1,28 @@
...
+ROOT_DIR = Path(__file__).parent.parent
+LOCAL_DIR = ROOT_DIR / "local"
+config = AutoConfig(ROOT_DIR)
+
+# Location of the Streetscapes data
+# ==================================================
+DATA_DIR = Path(
+    config("STREETSCAPES_DATA_PATH", LOCAL_DIR / "streetscapes-data")
+)
+DATA_DIR.mkdir(exist_ok=True, parents=True)
```
ClaireDons, line 18: We could replace this with checking for the csv files and throwing an error if they cannot be found.

notebooks/plot_city.ipynb
ClaireDons, line 1: We discussed adding a function to look if a file already exists for a city and if not to subset.

ClaireDons (changes requested):
Hey, I like how it's restructured. Just a few comments to improve usability and some of the issues I ran into when running it.

requirements.txt
cantordust: Nope! Removed...

pyproject.toml
cantordust: Done!

README.md
cantordust: Done!

README.md
cantordust: Indeed, I'll make those optional.

LICENCE.md
ClaireDons: Actually the licence is CC [CC by 4.0](https://creativecommons.org/licenses/by/4.0/)

streetscapes/conf.py
cantordust: I've cleaned up the configuration module, you shouldn't run into any more issues with that.

pyproject.toml
ClaireDons, line 1: I think it would be good to add ruff as a dependency so that we can format the files and do import sorting, linting and formatting more easily.

streetscapes/functions.py
ClaireDons, line 1: I checked running this function in a notebook in case a user in case a user prefers not using a CLI, which works well but it would be good if the prompt automatically defaulted to "y" if the code is run non-interactively. We can open an issue to implement this in the future though.  I also needed to explicitly give the data path in (but not when running it with CLI).

ClaireDons (commented):
Nice work, I was able to download all the images easily and things seem to run well.

I think it would be good to implement ruff for formatting and sorting import libraries as well as removing the dash lines between imports as that is more standard.   

For sorting libraries, we can run:
`ruff check --fix --select I`

PR #11, opened 2024-11-28
Initial work on image segmentation

cantordust:
- Segmentation pipeline using the [Mask2Former model](https://huggingface.co/docs/transformers/main/model_doc/mask2former). This is the model also used in the original `global-streetscapes` project. It supports 65 types of objects of varying granularity ranging from 'CCTV camera' to 'Mountain'. The `segment_from_dataset()` function takes in a data subset file (as generated, for example, with the `subset_data.ipynb` notebook) and segments all images specified within it. The assumption is that all the images have been downloaded (cf. `download_city_images.ipynb`).
- Segmentation statistics (median, mode, mean and standard deviation of the colour per category) are stored in an `xarray` Dataset and exported in NetCDF format.
- Visualisation function for displaying (partial) segmentation patches overlaid on the original image.
- A notebook demonstrating how to process a subset file and visualise the segmentation for a single image.

TODO
====

- Instance segmentation for more accurate colour estimation.


ClaireDons (approved):
Works great for me! I only had a quick look at the code since we all went through it yesterday. My only additional comment would be to add a little more context to the notebook about what we are looking at. That can be done in the future, when we work on documentation.

Peter9192:
BTW if it works for Claire then I would just merge it for now. I will continue testing later on and create new issues/PRs for anything I encounter.

ClaireDons:
Merging now as it works for me and an update addressing some comments is ready.

PR #13, opened 2024-12-02
Notebook warnings

cantordust:
A very small update that makes notebooks ignore warnings to avoid polluting the cell output with personal information.


ClaireDons (approved):
No problems running this!

PR #15, opened 2024-12-03
Panoptic segmentation

cantordust:
Better segmentation pipeline
- A separate submodule for segmentation models.
	- Added a base segmenter class with some common utility functions.
	- Added a wrapper class for the `Mask2Former` model. This can serve as a template for adding support for other segmentation models that can be used via a common API.

- Updated `pyproject.toml` with additional dependencies
- Updated the visualisation function. The legend is now nicely formatted.


ClaireDons (approved):
Seems to work well for me! I like the classes, I found it a lot easier to follow how the code works now. Only a few minor comments from me.

PR #18, opened 2024-12-05
Update README and dependencies

ClaireDons:
This updates the description of the project, installation instructions and adds badges as well as remove the custom mapillary dependency, so that we can publish to PyPI.


cantordust (approved):
Looks good!

PR #24, opened 2025-01-14
Explore data with duckdb

Peter9192:
I was missing an introductory notebook in the examples to familiarize myself with the streetscapes data. So I started adding one, building up to the amsterdam subsetting function. I used duckdb, and I find that it is superfast and may obfuscate the need for an intermediate parquet file.


Peter9192:
@ClaireDons please have a look to see what you think, you can start already while I'm still working on it.

Peter9192:
@ClaireDons if you could review the exploratory notebook, then I think you can create a new PR to continue working on the parquet conversion
@cantordust if you want to try out duckdb, you might also wanna check out this PR.

Suggestions to improve the SQL queries are very welcome :-)

ClaireDons (approved):
Great work! I rebuilt everything from scratch, including the docs and it all works well. Also impressed with how much faster duckdb is. No real comments on the code itself, other than there are couple of warnings that appear in the docs, which would be nice to get rid of. 

Should we create a new release for this, or should we wait for me to make a pull request for the Parquet files?

Peter9192:
> Should we create a new release for this, or should we wait for me to make a pull request for the Parquet files?

Let's wait for your changes as well

Peter9192:
Thanks for updating the parquet notebook!

PR #28, opened 2025-01-23
Combine more csvs into one file

ClaireDons:
This pull request builds upon #24 by combining more csvs into a parquet file as well as creating a more selective dataset that suits our needs.


Peter9192 (commented):
Thanks a lot for taking this further! It is almost complete, but I think there's a bug in combining multiple files (which I've probably introduced in my quick and dirty prototyping). Can you double-check?

docs/docs/examples/convert_csv_to_parquet.ipynb
```diff
@@ -66,15 +66,15 @@
...
+   "execution_count": 11,
    "metadata": {},
    "outputs": [
     {
      "name": "stdout",
      "output_type": "stream",
      "text": [
-      "human_readable(csv_size)='13.09 GB', human_readable(parquet_size)='3.39 GB'\n",
-      "reduction_factor=3.8614819275734855\n"
+      "human_readable(csv_size)='13.09 GB', human_readable(parquet_size)='5.24 GB'\n",
```
Peter9192, line 76: This size is calculated based on a glob and filtering out 'combined.csv'. But now that you renamed something to streetscapes.parquet, this file is also included in the statistic. Could you add that to the filter and re-run?

Peter9192 (approved):
Almost perfect! One more detail before we can merge this.

docs/docs/examples/convert_csv_to_parquet.ipynb
Peter9192: Or move the combined/streetscapes data to another dir?

ClaireDons: Should be sorted now! It doesn't seem to have changed my stats, I assume it's because they didn't exist when I ran the code the first time.

ClaireDons:
I also just wanted to note that when I was running the notebook to contribute to the global-streetscapes dataset (so on all csv files) that I had a type error come up. It could be solved by doing this:

`COPY (SELECT * FROM read_csv_auto('{file}', sample_size=-1))`

but it is significantly slower, so there might be better options.

Peter9192:
> I also just wanted to note that when I was running the notebook to contribute to the global-streetscapes dataset (so on all csv files) that I had a type error come up. It could be solved by doing this:
> 
> `COPY (SELECT * FROM read_csv_auto('{file}', sample_size=-1))`
> 
> but it is significantly slower, so there might be better options.

Sounds like your fix is scanning the entire file to infer the types of the columns. If you know which file it concerns, I suppose we could manually specify the type? But let's fix that afterwards.

docs/docs/examples/convert_csv_to_parquet.ipynb
Peter9192: I think it did, between your commits the reported reduction factor in the notebook first decreased to 2.5 and then back to 3.86

Peter9192:
Please go ahead and merge this

PR #33, opened 2025-02-20
Change paths to where data is stored

ClaireDons:
Changed paths where conf.OUTPUT_DIR is referenced to conf.DATA_DIR and the separate sub-directories so that the notebooks can be run properly. Also changed "Amsterdam-side.csv" to be a Parquet file.


cantordust (approved):
Looks good!

PR #36, opened 2025-02-21
Dinosam

cantordust:
Changelog

- Added a model called DinoSAM combining GroundingDINO + SAM2 for open-set instance segmentation with labels.
	- Insipred by the [LangSAM](https://github.com/luca-medeiros/lang-segment-anything) and [SamGeo](https://samgeo.gishub.org/) projects.
- Added functions to identify instances of objects specified as a hierarchy, which is used to remove overlaps across object modalities.
- Updated the dependencies in `pyproject.toml`.
- The DinoSAM model now has a unified API with the Mask2Former model.
- The metadata file now includes an instance segmentation mask.
- Removed the obsolete `conf.OUTPUT_DIR` variable and introduced `conf.CSV_DIR`, `conf.PARQUET_DIR` and `conf.IMAGE_DIR` for streamlined access to the correct directories.
	- The local directory structure mimics the remote HuggingFace repository structure for the Global Streetscapes data, at least for CSV and Parquet files
- Moved the default local data directory one level up (under `local/data` instead of `local/streetscapes-data/data`)
- Metadata and segmentations are now computed separately but saved together. This is a deliberate choice to avoid the two getting out of sync and introducing bugs that are hard to identify.
- The metadata is in the form of an `awkward` array. This allows for intuitive indexing even with ragged arrays. The `awkward` array is converted to JSON format when saving the metadata to a Parquet file and restored as an `awkward` array when loading the metadata from the file.
        - The metadata also includes longitude and latitude information for the images it refers to. Images without coordinates are skipped.
- Refactored the plotting method to handle the new metadata format.
- Example notebook demonstrating the same segmentation pipeline with the two models under `docs/examples/segmentation.ipynb`.
- Numerous other bug fixes and improvements.


src/streetscapes/models/dinosam.py
```diff
@@ -0,0 +1,382 @@
...
+        self.sam_model = SAM2ImagePredictor.from_pretrained(
+            self.sam_model_id, device=self.device
+        )
+
+        # GroundingDINO model.
+        # ==================================================
+        self.dino_processor = AutoProcessor.from_pretrained(self.dino_model_id)
+        self.dino_model = AutoModelForZeroShotObjectDetection.from_pretrained(
+            self.dino_model_id
+        )
```
Peter9192, line 102: I think you may have forgotten a `to_device` call here

Peter9192 (commented):
Nice, I was able to run the notebook after adding in a `to_device` call in the load method of DinoSAM. Haven't looked at all changes in detail, but overall it seems quite nice. 

Trying to process all images in 'Amsterdam_side' crashed my kernel. Do I get it correct that during segmentation/metadata extraction all results are kept in memory, until you call `save_metadata`? If so, how do we prevent hitting memory limits when analyzing big samples? 

I was a bit confused by the markdown cells reading:

> Those images will be segmented, and the metadata (statistics about the colour of the detected instances, instance masks, latitude and longitude of the place where the image was taken, and so forth) will be stored in a separate Parquet file for each image, with the same name as the image itself (and the extension `.parquet`) in the same directory as the image

Is that outdated? In the PR top post you seem to say something different, more in line with what I see in the code.

Finally, and this has not much to do with this PR specifically, but I don't like it much that we are loading amsterdam-side from a file. Many of the notebooks do this, but I keep running into issues with it where the file does not exist, or not in the right place, or whatever. I'd much prefer it if the `amsterdam-side` dataframe was generated on the fly from the original HF dataset using ibis/duckdb (possibly wrapped as a utility function in scs)

Peter9192:
Would be nice if `download_images` used `conf.IMAGE_DIR` as default path

Peter9192:
Would be nice if `segment_from_dataset` would accept a dataframe instance as dataset rather than only a file path

src/streetscapes/models/dinosam.py
cantordust: Thanks for the comments! This is still a draft PR, so work in progress, but I'm happy that you managed to run it - it's a good sign.🙂

The MD cells are indeed still a little out of order and not quite reflecting the code content in all cases. I was focusing on getting the functionality right. I'll clean up the MD text for the final PR.

Indeed, following our last discussion, the metadata is now a separate step, and it is up to the user. So the segmentation results have to be made accessible to the user somehow. One option is to save them to disk and load them when computing the metadata, but we would have to use some temporary storage (the local directory or `/tmp`), and then clean up after the metadata has been computed.

> Finally, and this has not much to do with this PR specifically, but I don't like it much that we are loading amsterdam-side from a file.  Many of the notebooks do this, but I keep running into issues with it where the file does not exist, or not in the right place, or whatever. I'd much prefer it if the amsterdam-side dataframe was generated on the fly from the original HF dataset using ibis/duckdb (possibly wrapped as a utility function in scs)

Sure, I can wrap that in a function, but that would add some overhead when running the notebook. It might be good to try to load the file first and then generate it if it's not there (that logic can be part of the function, of course).

cantordust:
That's something that I actually had planned, but I haven't got around to it yet. I'll implement it for the final PR.

Peter9192:
Need to fix conflicts - then merge and work on remaining tasks in follow-up PRs

src/streetscapes/models/dinosam.py
Peter9192: This issue (add `.to(self.device)`) is still present

Peter9192: Fixed in https://github.com/Urban-M4/streetscapes/pull/36/commits/39ff86ef6a6836994c96193a522f2133f4c1ad0c

Peter9192:
Had to add a `to(self.device)` call, but now I'm able to run it. DinoSAM is quite slow (3 images takes over 2 minutes), but results are impressive.

Since it all works, I will merge this now so we can continue with more atomic PRs :atom:

Peter9192 (approved):

PR #38, opened 2025-03-18
Add notebook exploring ZenSVI

Peter9192:
This notebook explores ZenSVI. CV module seems to work nicely. Downloading Amsterdam images a bit cumbersome.

Couldn't install ZenSVI as dependency in streetscapes due to environment conflicts, so I created a custom environment. Related CLI class in top cell of notebook. 

Cannot be merged unless issues are fixed. We could consider using the same/similar API structure with dedicated folders for images and their analysis, so as to ease integration later on.


cantordust:
All good, although installing ZenSVI was somewhat painful. Only Python 3.11 worked for me (I failed to run the notebook with any higher versions). Also, `python-decouple` was missing from the dependencies, so I would suggest adding that to the `!uv pip install zensvi ipykernel` line in the first cell.

Peter9192:
Closing as decided we're not going to incorporate ZenSVI in streetscapes directly

PR #42, opened 2025-03-28
Streetscapes workspaces

cantordust:
- Completely refactored the package and directory structure. We are n…ow fully harnessing the Huggingface facilities to download and cache data from the Global Streetscapes repository.
- Moving towards migrating to Ibis as a default for loading datasets (i.e., away from Pandas).
- Introduced workspace management for easy manipulation of different subsets of the larger Global Stretscapes dataset. This includes images, segmentations, instances and other files downloaded or generated by the user.
- Moved all functions to do with conversion, loading and saving of datasets into the Workspace class
- The API and the attributes are not completely settled...WIP


Peter9192 (commented):
Nice progress. Just a suggestion along the way: currently the workspace seems very much centered around the Global Streetscapes HF repo structure. This is nice, but in the future it would be nice if it could also work independently, e.g. download images from a specified bounding box directly from an image source (e.g. Amsterdam API). How do we accommodate these use cases? Is global streetscapes just a part of the complete workspace? Or a separate object altogether?

cantordust:
Yes, Global Streetscapes is just part of the workspace. It's easier and more predictable to manage the downloaded CSV and Parquet files with the Huggingface facilities. But images and other generated data will be stored in an asset directory managed separately for our package, and it will have separate subdirectories for different image sources.

docs/docs/examples/workspace.ipynb
ClaireDons, line 1: This works nicely for me, but it would be nice to have some basic description of how things work and also of the different options before we merge. For instance explaining that passing `create=True` creates a new workspace.

src/streetscapes/__init__.py
ClaireDons, line 1: Will these no longer be needed long-term?

src/streetscapes/info.py
ClaireDons, line 1: If this file is no longer needed, let's delete it?

ClaireDons (commented):
Hey, I like the new structure, it seems clearer to me than before. I only had a brief look since the repo still needs cleaning up. The notebook works well for me once, I knew to create a new workspace. I think before merging it would be good to add some description to the notebook to give context (and be able to run it for the first time easily).

PR #45, opened 2025-04-24
Docs update

ClaireDons:
Bringing the documentation up to date with current workflow

- [x] Update docs structure
- [x] Update mkdocs.yml 
- [x] Delete obsolete notebooks
- [x] Create overall tutorial
- [x] Replace global streetscapes with bbox retrieval in tutorial

The overall workspace tutorial should now work as well as the docs, with some changes that need to be addressed in future pull requests: 
- Calculating the stats and linking these to image ids so that it can be rasterised
- Most of the example notebooks are currently broken and need to be updated to the new method of working 
- Investigate the limit argument in the bbox retrieval (I currently set it to 500)


ClaireDons:
We went through this together and will merge now

PR #60, opened 2025-05-01
Create Global Streetscapes tutorial

ClaireDons:
A tutorial for using global streetscapes and exploring the dataset. I added two simple examples: perception of 'safe' on driving and walking surfaces and one where we calculate the building fraction and subset the data to only images that contain buildings. 

I also added a couple of methods to the global streetscapes source but they are clunky and should be streamlined. I'm not entirely sure how best to approach it though, so some feedback is welcome :)


Peter9192 (approved):
Looks nice. I noticed a few small things but I don't think we should spend too much time making this perfect.

- One of the first cells explains that we see files that we have already downloaded (with the tree view). But since this is the only remaining docs that use GSS, I think this would normally be (almost) empty. That might be confusing. Shouldn't we explicitly download "info.csv" first, for example?
- One of the markdown blocks near "exploring" mentions the use of `recreate=True`, but this is not reflected in the code cell below
- I think it might be nice to provide a little bit more general context at the top description. Something like "Global Streetscapes already analysed many of the images available on Mapillary and Kartaview. This rich metadata set could be a great starting point for our analysis as well. In this notebook, rather than fetching images directly from source, we start from the GS dataset to narrow down the images so we have a better starting point for our streetscapes analysis study".

PR #62, opened 2025-05-01
Data Sources tutorial

ClaireDons:


ClaireDons:
> Great to have up-to-date examples of all supported image input sources.
> 
> One comment is that I would suggest to make each "chapter" in the docs more self-contained/standalone. So instead of
> 
> > Now let's add which source of images we would like to use.
> 
> I would suggest something like:
> 
> > Streetscapes can work with images from multiple sources. Currently included are ..... Since they come from different providers, obtaining them works slighly differently for each source. Here we provide some brief examples of how to use each image source.

Thanks, I made it more general and also added a description of each data source and how each is different.

PR #63, opened 2025-05-08
Working with instances

cantordust:
The `SVWorkspace` class was updated and streamlined. Models and sources are not tied to a workspace any more and can be used independently. The workspace acts as a hub for managing generated data - for instance, loading and saving files for analysis.

The `SVSegmentation` and `SVInstance` classes were also updated, with some useful methods added for extracting individual channels, merging multiple instances with the same label and visualising instances in isolation.


ClaireDons (approved):
Nice work, I think it works nicely now :) I made a couple of changes to the tutorial notebook to fix some broken links and add more context to the statistics calculation. I also fixed the docs not being able build. I think the tutorial is in working condition now though! 

A couple of other things I noticed is:
- The progress bar for the segmentation is a but misleading, it shows 100% done but then it continues to run for longer. (I assume there is another process going on after the segmentation that also takes time?)
- When  you process a lot of images, there is a chance that some images don't contain a building which crashes calculating the mean. For now I included an if statement but it's a little clunky.

PR #68, opened 2025-05-13
Segmentation archive format and workspace metadata

cantordust:
- Changed the format for segmentations to .npz (NumPy archive)
- Initial code for saving and loading workspace metadata


ClaireDons (approved):
This works well for me and is definitely a lot faster :) I think the metadata bit needs a bit more detail but that can be bundled up in a PR for the tutorial fixes later.

PR #75, opened 2025-07-29
Simplify workspace

Peter9192:
This PR intends to make the behaviour of the workspace and mapillary class more predictable, so that it is easier to understand what's going on for new users. It is informed by the Amsterdam usecase as set out in #74.

- `env` object is quite confusing, stop passing it around
- remove huggingface cache management in favour of user-managed directories
- remove metadata object on workspace; add load/save metadata functions instead
- refactor mapillary fetch metadata to simplify storing it in workspace

TODO: 
- update notebooks
- make it work on snellius


Peter9192:
We just concluded that a sample of 10_000 images (just using `sample` on the complete metadata) gives a nice even distribution of images throughout Amsterdam. TODO: redo download images but filtered by creator/is_pano and start with a sample of 1000.

Peter9192:
Opened #76 to wrap up work originally started in this PR

ClaireDons (approved):
Approving this now to fix any issues later in separate pull requests.

PR #80, opened 2025-09-19
Make Mapillary great again

Peter9192:
Rework the mapillary module and CLI to fix downloading images from mapillary

fetch metadata
- [x] works for a simple CLI invocation
- [x] has example/test to verify it keeps working
- [x] exports manifest that can be read again
- [x] works at scale
- [ ] ~can resume if interrupted~ seems quite stable atm, but can easily be added back later

download_images
- [x] works for simple CLI invocation
- [x] has example/test to verify it keeps working
- [x] exports manifest that can be read again (and is similar to the previous one)
- [ ] works at scale
- [x] can resume if interrupted


Peter9192:
TODO: 
- streamline duckdb cache system between metadata and downloader; 
- add more logging to downloader
- fix logging and progress bar combi (also use tqdm as before? don't like the rich one)
- add (more) testing (for various options for input e.g. default manifest dir for downloading vs specify explicitly)
- ...

Peter9192:
Current directory approach is flawed because it will restart at 0 when starting a new download

Peter9192:
Quite happy with the current state of this (always more to do, but can be done later). Perhaps streamlining the docstrings still. But a (superficial) review of the mapillary/config/project/db behavour might be nice @cantordust

README.md
```diff
@@ -171,6 +171,29 @@ df_result = agg_expr.execute()
...
+## Recommended Output Directory Structure
+
+All CLI and API outputs are stored under a configurable base directory (default: `output/`). You can set the base directory via `.env` (e.g., `DATA_HOME=output`) or CLI options.
+
+Example structure:
+
+```
+output/
+  manifests/           # All metadata manifests (GeoParquet, CSV, etc.)
+  images/              # All raw images (optionally flat, or sharded for scale)
```
cantordust, line 183: Should images be considered 'input' instead of 'output'? If you want to start fresh, you should be able to just wipe the output directory, but that would also remove the images you've downloaded.🙂

src/streetscapes/config.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+def get(key: str, default=None):
+    return load().get(key, default)
+
+
+def set(key: str, value):
+    cfg = load()
+    cfg[key] = value
+    CONFIG_FILE.write_text(json.dumps(cfg, indent=2))
```
cantordust: Nice, I think this is going in the right direction! I like the persistence aspect of it. 

A couple of suggestions and comments:

1. [`OmegaConf`](https://omegaconf.readthedocs.io/en/2.3_branch/#) might be a good option here. The format is YAML rather than JSON, but on the flipside it has nice functionality such as easy nested access (e.g., via attribute-like strings, such as `config.models.dino.confidence = 0.5`). You also get `pydantic`-style type safety if you use data classes for the configuration.
2. It might be good to use cross-platform destinations via [`platformdirs`](https://github.com/tox-dev/platformdirs) - e.g., `platformdirs.user_config_dir()`.
3. Like the `console` object above, you can expose a potential `OmegaConf` object via `__init__.py`.

src/streetscapes/cli/console.py
```diff
@@ -0,0 +1,12 @@
+import logging
+
+from rich.console import Console
+from rich.logging import RichHandler
+
+console = Console()
```
cantordust, line 6: Just an idea: expose the `console` object via `__init__.py`.

src/streetscapes/cli/download_images.py
```diff
@@ -1,43 +1,106 @@
...
+           11:     149mm x 149mm
+           12:   37.2mm x 18.6mm
+        Each level of precision subdivides the previous level into 32 subtiles.
+    Shard path of precision 7, split in three parts abc/de/fg
+    abc/ --> region level
+    de/ --> neighbourhood scale (max 32x32 = 1024 per region)
+    fg/ --> block level  (max 32x32 = 1024 per neighbourhood)
+    """
+    geom = from_wkb(geometry)
+    geohash = pygeohash.encode(geom.y, geom.x, precision=7)  # 153m x 153m
```
cantordust, line 105: Nice, I like `pygeohash` as a concept. One question - why is the precision hardcoded as `7`? Is there any value in giving the user the ability to specify the precision manually?

src/streetscapes/cli/export_table.py
```diff
@@ -0,0 +1,36 @@
...
+    elif ext in [".parquet"]:
+        project.export_parquet(table_name, output)
+    elif ext in [".json"]:
+        project.export_json(table_name, output)
+    elif ext in [".gpkg"]:
+        project.export_gpkg(table_name, output)
+    elif ext in [".geojson"]:
+        project.export_geojson(table_name, output)
+    else:
+        raise typer.BadParameter(f"Unsupported file extension: {ext}")
```
cantordust: Suggestion: You could use a LUT (dictionary mapping extension to function, since the signatures are the same):

```python
exporters = {
    ".csv": project.export_csv,
    ".parquet": project.export_parquet,
    ...
}

if exporter := exporters.get(ext) is None:
    raise typer.BadParameter(f"Unsupported file extension: {ext}")

exporter(table_name, output)

```

src/streetscapes/cli/segment_images.py
```diff
@@ -60,7 +59,7 @@ def _save_masks(masks, out_dir: str, filename: str):
 # Standalone models that can operate directly on a list of images
 # ---------------------------
 @segment_images_cli.command("sam")
-def segment_images_sam(images: list[str], out: str):
+def segment_images_sam(images: str, out: str):
     """Segment images with SAM."""
     images_list = _parse_image_input(images)
     model = SAM(checkpoint="sam_vit_h_4b8939.pth")
```
cantordust, line 65: Suggestion: Maybe an option to specify the checkpoint - you might want to use a custom fine-tuned model. I can add that when I mould the model API to this style.

src/streetscapes/config.py
```diff
@@ -0,0 +1,37 @@
+import json
+from pathlib import Path
+
+CONFIG_FILE = Path.home() / ".config/streetscapes/config.json"
+CONFIG_FILE.parent.mkdir(parents=True, exist_ok=True)
+
+DEFAULTS = {
+    "data_home": str(Path.home() / ".local/share/streetscapes"),
+    "active_project": "streetscapes",
+}
```
cantordust: Suggestion: dataclasses for type safety (also cf. the comment below).

src/streetscapes/models/base.py
```diff
@@ -169,14 +163,14 @@ def _flatten_labels(
             A flattened category tree where each key is a
             category and the corresponding value is a list of
             masks that should be subtracted from it.
+
         """
 
         def _flatten(
             tree: dict,
             _subtree: dict = None,
         ) -> dict:
```
cantordust, line 172: @cantordust Note to self: this is unsightly, needs to be reworked into a more intuitive representation.

src/streetscapes/project.py
```diff
@@ -0,0 +1,149 @@
+from pathlib import Path
+
+import ibis
+from shapely.geometry import box
+
```
cantordust, line 6:
```suggestion
from pandas import DataFrame
```

src/streetscapes/project.py
```diff
@@ -0,0 +1,149 @@
...
+    def image_dir(self, source: str | None = None):
+        if source is None:
+            return self.data_home / "images"
+        return self.data_home / "images" / source
+
+    def get_table(self, name: str):
+        """Return an ibis table reference."""
+        return self.con.table(name)
+
+    def ingest_mapillary(self, df, table_name="mapillary"):
```
cantordust:
```suggestion
    def ingest_mapillary(self, df: DataFrame, table_name="mapillary"):
```

src/streetscapes/project.py
```diff
@@ -0,0 +1,149 @@
...
+            # TODO: consider configurable duplicate behaviour (REPLACE or IGNORE)
+            self.con.raw_sql(f"""
+                INSERT OR REPLACE INTO {table_name}
+                SELECT
+                    * EXCLUDE (geometry),
+                    ST_GeomFromText(geometry) AS geometry,
+                FROM metadata_tile
+            """)
+
+    def filter_bbox(self, table_name, bbox):
```
cantordust: (Also import `Bbox` from `fetch_metadata.py` if that's the right type)

```suggestion
    def filter_bbox(self, table_name: str, bbox: Bbox):
```

src/streetscapes/project.py
```diff
@@ -0,0 +1,149 @@
...
+
+    def __init__(self, name: str = "streetscapes"):
+        # TODO: also read name from config? But keep option to overwrite?
+        self.name = name
+        self.data_home = Path(config.get("data_home"))
+
+        database_path = self.data_home / "projects" / f"{name}.duckdb"
+        database_path.parent.mkdir(parents=True, exist_ok=True)
+        self.con = ibis.duckdb.connect(database_path)
+        self.con.raw_sql("INSTALL spatial; LOAD spatial;")
```
cantordust, line 22:
```suggestion
        self.con.con.install_extension("spatial")
        self.con.con.load_extension("spatial")
```

src/streetscapes/sources/mapillary.py
```diff
@@ -1,286 +1,211 @@
...
-
-    Methods:
-        init(path: Path) -> None:
-            Initialize storage (e.g., create empty file if missing).
-        append(records: List[dict], path: Path) -> None:
-            Append a batch of records to storage.
-        read(path: Path) -> Any:
-            Read the entire stored dataset.
-    """
+Bbox = tuple[float, float, float, float]
```
cantordust: This is also defined in `fetch_metadata.py`.

src/streetscapes/streetview/imageset.py
```diff
@@ -1,3 +1,4 @@
 class SVImageSet:
     """TODO: Add docstrings"""
+
     pass
```
cantordust, line 4: I'm not sure if this class makes sense. You could just use containers (lists or dictionaries) of `SVImage`s and use `map` to apply any functions to all objects 🤔.

src/streetscapes/streetview/workspace.py
cantordust, line 1: Is the workspace paradigm obsolete (or redundant?) now that you have the `Project` class?

src/streetscapes/utils/bbox.py
```diff
@@ -0,0 +1,42 @@
+from itertools import product
+from typing import Iterable
+
+import numpy as np
+
+Bbox = tuple[float, float, float, float]
```
cantordust, line 6: Maybe just import this into `fetch_metadata.py` and `mapillary.py` (or expose it via `__init__.py` and import it from there).

Also, would it make sense for `Bbox` to be a (simple) class (erhaps a `dataclass`)?

tests/conftest.py
```diff
@@ -0,0 +1,64 @@
...
+                "geometry": {"coordinates": [4.9, 52.37]},
+                "computed_geometry": {"coordinates": [4.9, 52.37]},
+                "captured_at": "2023-01-01T00:00:00Z",
+                "thumb_2048_url": "https://example.com/image1.jpg",
+            }
+        ]
+
+    def fake_download_image(self, url, path):
+        path.parent.mkdir(parents=True, exist_ok=True)
+        path.write_text("FAKE IMAGE DATA")  # simulate an image file
```
cantordust, line 57: Should this be an actual image rather than text? You might want to use the image in other tests. Perhaps something like `skimage.data.astronaut()`?

cantordust (approved):
Great refactoring effort🎉! Some raw SQL could probably be replaced with Ibis (or even native DuckDB) methods, and perhaps some further refactoring would be beneficial to remove redundancies like the multiple `Bbox` definitions. Also, I would like to get a better understanding of the difference between workspace and project. But at any rate this is a nice new base that the model API can be built on!

README.md
Peter9192: Ahh I think the README is already outdated. Let's fix that in a follow-up PR. Maybe after also adding back some segmentation functionality.

src/streetscapes/config.py
Peter9192: Thanks for the suggestions. Plaformdirs is nice, omegaconf as well. I've added both, except that I'm not exposing the omegaconf object directly, but still through the custom getter/setter. This is to make sure all settings are persisted. We could change that in the future, but I want to wrap this up now.

I'm not sure I like re-exporting everything in __init__.py for shorter imports.

src/streetscapes/cli/console.py
Peter9192: I'm not sure I like that. Let's leave it as is for now, and decide later on if/when it makes sense to do this.

src/streetscapes/cli/download_images.py
Peter9192: I think the 153 m resolution makes sense for streetscapes applications. This is not typically something you'd want to change midway through the project, because your folder structure depends on it. But for the user-facing API it might make sense. We could consider that later on. E.g. store it in the config

src/streetscapes/cli/export_table.py
Peter9192: Yes, I agree. Nice example of the walrus operator btw 😉

src/streetscapes/sources/mapillary.py
Peter9192: Historical reasons. Can be removed indeed.

src/streetscapes/streetview/imageset.py
Peter9192: That's outdated I think. At least irrelevant for this PR. Can be discarded later.

src/streetscapes/streetview/workspace.py
Peter9192: It's one or the other, I haven't decided which name I like best. Currently `Project` is the one to keep.

src/streetscapes/utils/bbox.py
Peter9192: I had that before (or a namedtuple), but that doesn't play well with typer's type annotations. Maybe if we switch to cyclopts?

Peter9192: And indeed, I now import this one in the other modules.

tests/conftest.py
Peter9192: Nice idea, let's do that in the future once we really want to test something about the image.

src/streetscapes/cli/segment_images.py
Peter9192: Yes, let's add that later.

src/streetscapes/config.py
Peter9192: I kept it super simple for now, but that would be an option indeed. I want to avoid switching to something like pydantic, that's too heavy handed in my opinion.

src/streetscapes/models/base.py
Peter9192: I think we could just use a flat list, right? Do merging/subtraction in the postprocessing/analysis.

src/streetscapes/project.py
Peter9192: I don't have a strong preference for either of these.

PR #81, opened 2025-11-04
Model CLI

cantordust:
## A servable model pipeline [WIP]

This is designed to facilitate CLI-based workflows. The idea is that models should not be loaded with every CLI command call, but rather spawned once and queried repeatedly. 

A demo CLI entry point is provided for the `MaskFormer` model. It can be called with 

`streetscapes segment_images maskformer <image_path>`

It will check if the model service is alive and spawn it if necessary. **This should not happen inside the entry point**. Instead, there should be a series of `streetscapes model serve` commands for each model. However, right now this is OK to ensure that the pipeline works.


src/streetscapes/serve/server.py
```diff
@@ -0,0 +1,60 @@
...
+from streetscapes.models.maskformer import MaskFormer
+
+os.environ["RAY_ACCEL_ENV_VAR_OVERRIDE_ON_ZERO"] = "0"
+
+
+@serve.deployment(
+    num_replicas=1,
+    ray_actor_options={
+        # TODO: These should be taken from the configuration.
+        "num_cpus": 0.2,
```
Peter9192, line 20: I guess we should add the configuration to our todo list issue.

src/streetscapes/sources/downloader.py
```diff
@@ -0,0 +1,111 @@
+import hashlib
+import shutil
+from pathlib import Path
+import datetime
+from rich.progress import track
+
+
+class ImageDownloader:
```
Peter9192, line 8: This seems outdated. CLI for downloading images is added in https://github.com/Urban-M4/streetscapes/pull/80

I guess we should clean this up (is on the TODO list). Agreed?

src/streetscapes/models/maskformer/functions.py
```diff
@@ -0,0 +1,174 @@
...
+                    "image": oj.dumps(image, option=oj.OPT_SERIALIZE_NUMPY),
+                }
+                for h, image in zip(hashes, images)
+            ],
+            "labels": labels,
+        }
+        response = handle.remote(data).result()
+
+        # Store the segmentations and their metadata
+        save_segmentations(project, model_params, response, to_overwrite)
```
Peter9192, line 174: loading/saving the segmentation is not the responsibility of the model, right? I wonder if we envision moving this more centrally in the CLI/project/db modules

src/streetscapes/models/maskformer/functions.py
```diff
@@ -0,0 +1,174 @@
...
+                .to_pydict()["uuid"][0]
+            )
+        else:
+            seg_uuid = uuid.uuid4()
+
+        seg_fpath = project.get_output_dir("maskformer", True) / f"{seg_uuid}.npz"
+
+        # Save the segmentations.
+        np.savez(seg_fpath, instances=segmentation.instances, masks=segmentation.masks)
+
```
Peter9192, line 77: I'm trying to make sense of the output file, but just reading the numpy file isn't very intuitive: 

```
data = np.load('...520e63403399.npz', allow_pickle=True)
data['instances']  # okay, that I can work with
data['masks'] # ...... np.int32(19), np.int32(19), np.int32(19), np.int32(19), np.int32(19), np.int32(19), np.int32(19), np.int32(19), np.int32(19), np.int32(19), np.int32(19)]},
      dtype=object)
      
```

that seems like an enormous list of np.ints, seems inefficient and hard to work with. 

I liked it a lot when I had something like:

```py
data[masks] = [np.array([[...]], dtype=bool), np.array([[...]], dtype=bool), ...]  # list of 2d arrays of booleans
data[labels] = ['person', 'building', 'building', ... ]  # list of labels corresponding to each mask

Peter9192 (approved):
Thanks, this now seems to work pretty well. I was able to run on cpu/gpu after playing around with ray serve config and batch size. 

My main concern is that I'd like to agree on a different output format for the segmentations. Other than that I have some small comments. All of that can be addressed later, happy to see this merged as is.

src/streetscapes/sources/downloader.py
cantordust: Aye, aye!

src/streetscapes/models/maskformer/functions.py
cantordust: I agree that saving the output should not be up to the model, but we have to consider that each model will output something different, so it might be tricky to make a unified output saving function.

src/streetscapes/models/maskformer/functions.py
cantordust: I'm on board with that, it's part of the TODO list ✅.

PR #83, opened 2025-12-09
Adjust visualization script to adjusted output format

Peter9192:
This adjusts the SVSegmentation object to be compatible with the output structure from https://github.com/Urban-M4/streetscapes/blob/600f28180575617360b334c371b44d559396f91f/scripts/segment_optimized.py#L246-L254

to try it out, re-install the package from this branch:

```bash
pip install git+https://github.com/Urban-M4/streetscapes.git@fix-visualize
```


cantordust:
@Peter9192 Do we still need this?

BSchilperoort:
Outdated

PR #86, opened 2025-12-19
Model intermediate table

cantordust:
Intermediate table structure for filtering segmentations by model. This may evolve to incorporate the subset definition.


src/streetscapes/models/maskformer/functions.py
```diff
@@ -141,23 +105,27 @@ def segment_images(
...
     if model_params is None:
         model_params = {}
 
     if labels is None:
         labels = {l: None for l in MaskFormer.id_to_label.values()}
 
-    to_process, to_overwrite = get_unprocessed_images(project, image_paths, overwrite)
+    (processed, unprocessed) = project.get_image_status(image_paths, "maskformer")
+    if overwrite:
+        processed = {}
```
Peter9192: Don't you need to add the processed images to `unprocessed` then?

cantordust: Yeah, that was a leftover from a previous attempt at filtering the processed images. Fixed now - processing the same image again would result in the same UUID being used for the segmentation.

PR #87, opened 2025-12-19
Add BFMS model using ray serve

Peter9192:
This adds the BFMS model

* I moved the `async def process` call to a dedicated wrapper function in `service.py` for both maskformer and BFMS. This keeps the model agnostic about the ray serve deployment - better separation of concerns.
* I changed `server.py` to make use of these new service wrappers
* I hit some circular imports, so removed re-exports in `__init__.py` files
* I placed the CLI in the model's module so cyclopts does lazy imports - if we like it, also do this for maskformer?
* I removed the `segment_images` function in `functions.py` because I felt it was doing orchestration that didn't belong to the 'segmentation'. 
* I removed batching for simplicity at first. I wonder if we could use ray serve automatic batching
* I 

# TODO
* Reimplement save_segmentation (both on disk and in database).
* Take images from subset/database instead of path


Peter9192:
@cantordust this modifies quite a bit of maskformer logic. Would be good if you could have a look and we can sync on what to keep.

src/streetscapes/models/bfms/cli.py
```diff
@@ -0,0 +1,62 @@
...
+    """
+    # Resolve paths
+    image_path = Path(image_path)
+    if image_path.is_dir():
+        image_paths = [p for p in image_path.glob("*.*") if ft.is_image(p)]
+    else:
+        image_paths = [image_path]
+
+    if not image_paths:
+        return
```
cantordust, line 38: Just a suggestion, we are going to have a few of these entry points with the same kind of preprocessing logic, so maybe it makes sense for this to be a dedicated function in the `utils` module.

src/streetscapes/models/bfms/db.py
```diff
@@ -0,0 +1,59 @@
...
+
+    Args:
+        project: Current project.
+        params: The model parameters.
+        segmentations: The list of segmentations.
+        processed: Images that have already been processed.
+    """
+
+    # Rows to be inserted into the database
+    seg_rows = {k: [] for k in SCHEMA()}
```
cantordust, line 34: Do you need the `()` in `SCHEMA()`?

src/streetscapes/models/bfms/model.py
```diff
@@ -1,53 +1,69 @@
...
+        "url": "https://figshare.com/ndownloader/files/50246925?private_link=fd38d547fdb8708381f5",
+        "md5": "1b32428cfb4f6cfff8800779364289d4",
+    },
+    "model.safetensors": {
+        "url": "https://figshare.com/ndownloader/files/50246928?private_link=fd38d547fdb8708381f5",
+        "md5": "30c3999e43d1ee20c0685e92256c5d11",
+    },
+}
+
+MODEL_PATH = Path(user_data_dir("streetscapes"), "models", "bfms")
```
cantordust:
```suggestion
MODEL_PATH = user_data_path("streetscapes") / "models/bfms"
```

src/streetscapes/models/bfms/model.py
```diff
@@ -1,53 +1,69 @@
...
 import numpy as np
+import orjson as oj
 import torch
 from PIL import Image
-from transformers import (
-    AutoImageProcessor,
-    Mask2FormerConfig,
-    Mask2FormerForUniversalSegmentation,
-)
+from platformdirs import user_data_dir
```
cantordust, line 9:
```suggestion
from platformdirs import user_data_path
```

cantordust (approved):

src/streetscapes/models/bfms/cli.py
Peter9192: I agree, let's apply the rule of three here and create that utility when it's added the third time

PR #90, opened 2026-01-21
Add (mostly mocked) fastapi implementation

BSchilperoort:
For dev mode do:
    `fastapi run src/streetscapes/fastapi/server.py`

For end-user mode do:
    `streetscapes-explorer`

Note: this will require reinstalling the package, and for fastapi do `pip install fastapi[standard]` (pyproject.toml deps have been updated to reflect this).


src/streetscapes/explorer/server.py
```diff
@@ -0,0 +1,139 @@
...
+
+
+@app.post("/images/{image_id}/{model_name}/{instance_idx}/{label}")
+async def set_instance_label(
+    image_id: str, model_name: str, instance_idx: int, label: str
+):
+    """Set the label of a specific instance within a segmentation."""
+    pass
+
+@app.post("/images/{image_id}/segment/{model}/{run_args}")
```
BSchilperoort, line 138: These args for the image segmentation look OK, @sverhoeven ?

Also, do we need a separate method for multi-image segmentation, or should this allow for multiple image id's?

pyproject.toml
```diff
@@ -72,8 +73,13 @@ lint.select = ["F", "E", "TYP", "B", "D", "C", "Q", "I"]
 [tool.ruff.lint.pydocstyle]
 convention = "google"
 
-[dependency-groups]
+[project.optional-dependencies]
```
BSchilperoort, line 77: Made it so dev dependencies can be installed with `pip`.

I think the idea of the `[dependency-groups]` section was so you can install only dev dependencies? However `uv sync --only-group dev` still needs to resolve the heavy ML dependencies :shrug:
So  I'm not sure what the point of this is.

src/streetscapes/explorer/server.py
```diff
@@ -0,0 +1,171 @@
...
+    allow_headers=["*"],
+)
+
+
+def _inbounds(img: Image | ImageMetadata, bbox: Bbox) -> bool:
+    # Temporary implementation.
+    # Full implementation needs to account for spherical coordinates properly
+    if bbox.n > img.lat > bbox.s and bbox.w > img.lon > bbox.e:
+        return True
+    return False
```
cantordust:
```suggestion
    return bbox.n > img.lat > bbox.s and bbox.w > img.lon > bbox.e
```

src/streetscapes/explorer/server.py
cantordust: For multiple images, we have the concept of a `collection` - perhaps that should be part of the API. But then it depends on what granularity that we should support. If we want to be able to segment images individually (and adjust the model parameters **per image**), then the `collection` would be irrelevant for segmentation.

src/streetscapes/explorer/server.py
```diff
@@ -0,0 +1,171 @@
...
+async def _serve():
+    server = _start_uvicorn()
+    print("Waiting for the streetscapes-explorer to start...")
+    await asyncio.sleep(5)
+    webbrowser.open(
+        "https://urban-m4.github.io/Urban-M5/?s=http://localhost:5000"
+    )
+    print(
+        "The streetscapes-explorer should have launched automatically.\n"
+        "To start it manually, go to https://urban-m4.github.io/Urban-M5/ and "
```
cantordust:
```suggestion
        "To open it manually, go to https://urban-m4.github.io/Urban-M5/ and "
```

src/streetscapes/explorer/server.py
```diff
@@ -0,0 +1,171 @@
...
+        "paste in https://0.0.0.0:5000 as web service.\n"
+        "You will need to disable your ad blocker (like uBlock Origin Lite)"
+        " and allow your web browser to load localhost resources."
```
cantordust, line 162:
```suggestion
        "and allow your web browser to load localhost resources."
```

src/streetscapes/explorer/server.py
```diff
@@ -0,0 +1,171 @@
...
+    await server.serve()
+
+
+async def _serve():
+    server = _start_uvicorn()
+    print("Waiting for the streetscapes-explorer to start...")
+    await asyncio.sleep(5)
+    webbrowser.open(
+        "https://urban-m4.github.io/Urban-M5/?s=http://localhost:5000"
+    )
```
cantordust, line 156: Nice!🙂

cantordust (approved):

src/streetscapes/explorer/server.py
BSchilperoort: space is needed due to string concatenation

src/streetscapes/explorer/server.py
BSchilperoort: The front-end currently does not know of collections AFAIK.

For this POST method collections don't matter yet as it's about adjusting a single image's segmentation (if the user draws or relabels something).

PR #118, opened 2026-02-23
Add proper cli, allow port config for explorer

### LINKED ISSUE #108: Allow port configuration, change default port
Currently users cannot set the exporer's port through the cli. They should be able to do `streetscapes-explorer --port 9999`

Also, port 5000 seemed to be in use already on MacOS and snellius. A [different default is better](https://en.wikipedia.org/wiki/List_of_TCP_and_UDP_port_numbers).

BSchilperoort:
Closes #108 

Can still change default port #.

Remember to reinstall the package due to the change in the pyproject.toml file.


src/streetscapes/explorer/server.py
```diff
@@ -268,29 +270,61 @@ async def segment_image(image_id, model, run_args):
...
 
 
-def serve():
-    """Start the Streetscapes Explorer server."""
-    asyncio.run(_serve())
+cli = App(help="Streetscapes data explorer")
+
+@cli.default
+async def serve(
+    *,
```
BSchilperoort, line 318: @cantordust this is how you can configure cyclopts to not use the arguments as possible environment vars

E.g.;

```
Usage: streetscapes download_images mapillary [ARGS]

Download Mapillary images to a local directory.

╭─ Parameters ────────────────────────────────────────────────────────────────────────────────────╮
│ SKIP-EXISTING --skip-existing  If true, only download missing images; otherwise overwrite.      │
│   --no-skip-existing           [default: True]                                                  │
│ TOKEN --token                  Mapillary OAuth token (if not set via MAPILLARY_TOKEN).          │
│ PROJECT --project              An optional project to attach to.                                │
╰─────────────────────────────────────────────────────────────────────────────────────────────────╯
```
the `BOLD` arguments mean that they can also be passed as env var.  I don't think that's desirable here (maybe except for the token one, but that's not how it's used here).

src/streetscapes/explorer/server.py
```diff
@@ -268,29 +270,61 @@ async def segment_image(image_id, model, run_args):
     pass
 
 
-async def _start_uvicorn():
-    config = uvicorn.Config(app, host="0.0.0.0", port=5000, log_level="info")
+async def _start_uvicorn(port: int, log_info: bool):
+    config = uvicorn.Config(
+        app,
+        host="0.0.0.0",
```
cantordust: A nitpick, but should this be `127.0.0.1` by default?

src/streetscapes/explorer/server.py
```diff
@@ -278,27 +280,61 @@ async def segment_image(image_id, model, run_args):
...
+        port=port,
```
cantordust: Just a thought: maybe the host and port can also go into the configuration.

src/streetscapes/explorer/server.py
```diff
@@ -31,6 +32,7 @@
 
 origins = [
     "http://localhost",
+    "https://localhost",
     "http://localhost:5173",
```
cantordust: What is running on port 5173?

cantordust (approved):
Works with the updated configuration and picks up the port number when passed as a CLI argument.

src/streetscapes/explorer/server.py
BSchilperoort: Stefan had added this not sure why

src/streetscapes/explorer/server.py
BSchilperoort: not sure which is better here. If it's set to 127.0.0.1 only the local machine can reach the server. While 0.0.0.0 would allow any machine that can communicate with the host to reach the explorer. Not sure if an ssh tunnel to snellius is treated as a local machine.

src/streetscapes/explorer/server.py
BSchilperoort: Could add the port I guess. Not sure if we need to set the host if we leave it at 0.0.0.0

src/streetscapes/explorer/server.py
BSchilperoort: his frontend locally running needed that. Set it now to * so any origin is fine

src/streetscapes/explorer/server.py
BSchilperoort: see #130

PR #120, opened 2026-02-24
Convert dinosam instances to polygons

BSchilperoort:
Conversion works fine. However, results are not very nice. The segmentation masks are very blotchy and of low quality.

image

image


src/streetscapes/serve/server.py
```diff
@@ -45,6 +47,9 @@ def get_model_app(model: str, /, **kwargs) -> serve.Application:
...
 
 def serve_model(model: str, /, **kwargs) -> DeploymentHandle:
-
     app = get_model_app(model, **kwargs)
-    return serve.run(app)
+
+    ray.init(log_to_driver=False)
+    logger = logging.getLogger("ray.serve")
+    logger.setLevel(logging.WARNING)
+    return serve.run(app, logging_config={"log_level": logging.WARNING})
```
BSchilperoort, line 55: Kinda ridiculous but this seems to be the only way to make ray more quiet and not spam the console as much with info that's not useful to the end-user.

PR #121, opened 2026-02-26
Grab BFMS from huggingface, convert to polys

### LINKED ISSUE #119: Fix BFMS segmentation
The BFMS model was automatically downloaded from figshare. However, figshare has since blocked any anonymous programmatic access.

We can switch to the huggingface version of the model https://huggingface.co/jinfengxie/BFMS_1014
It also has a nice app on huggingface; https://huggingface.co/spaces/jinfengxie/BFMS_Mask2Former (maybe add to docs?)

To-dos;

- [x] Assert results of huggingface version match figshare version
- [x] Modify code to serve model from huggingface

BSchilperoort:
Fixes #119 

Tested if the results for the HF and Figshare models are the same, seems idential;

```py
>>> import numpy as np
>>> figshare = np.load("/home/bart/.local/share/streetscapes/archives/bfms/bfms-seg-figshare/u1/hp/8z/6e002a2f-9c3f-9c14-c93d-532c940c5b37.npz")["arr_0"]
>>> hf = np.load("/home/bart/.local/share/streetscapes/archives/bfms/bfms-seg/u1/hp/8z/6e002a2f-9c3f-9c14-c93d-532c940c5b37.npz")["arr_0"]
>>> np.all(figshare == hf)
np.True_
```

Also tried for a set of 10 segmented files, they're all identical.


BSchilperoort:
BFMS polygon conversion etc. works now. #119 exists to track bad results

PR #126, opened 2026-02-26
Refactor config

cantordust:
Added a default `project_dir` entry to the configuration where project databases and segmentation data are saved. This is different from the `image_dir` directory, which contains the downloaded images. The rationale for separating the two is that the image cache can be purged without affecting the databases and segmentations.

Also moved config and validation to Pydantic. **Old configs and setups are not valid anymore**.


src/streetscapes/config.py
BSchilperoort: While were changing the config anyway, can we add a check here that validates the config entry someone is trying to set? I.e. check that `key` is in `valid_keys`.

For example, if someone misspells a config entry it's not stored, and the user is returned a clear error message.

src/streetscapes/project.py
```diff
@@ -98,7 +98,7 @@ def __init__(
         self.root_dir = utils.ensure_dir(
             config.get(
                 "root_dir",
-                root_dir or pdirs.user_data_path("streetscapes"),
+                data_dir or pdirs.user_data_path("streetscapes"),
```
BSchilperoort: self.root_dir is data dir?

src/streetscapes/project.py
```diff
@@ -110,7 +110,7 @@ def __init__(
             config.get(
                 "data_home",
```
BSchilperoort: Also still not correct here

src/streetscapes/config.py
cantordust: Sure, but this means that we have to have a default for every allowed key.

src/streetscapes/project.py
cantordust: Yeah, all of those need to be refactored.

src/streetscapes/config.py
BSchilperoort: No? Just check that it is a key that can be set right

src/streetscapes/project.py
cantordust: No, it was meant to be the directory containing the `projects` and `archives` directories. I removed it since we are not going to have separate files for model outputs (which were saved under `archives`).

src/streetscapes/config.py
cantordust: But how do we know what keys are allowed? I thought one way would be to add them to the `DEFAULTS` dictionary, but I'm open to suggestions.☺️

src/streetscapes/explorer/server.py
```diff
@@ -43,16 +43,18 @@
     allow_headers=["*"],
 )
 
-dbpath = Path(
-    f"{config.get("data_home")}/projects/{config.get("active_project")}.duckdb"
-)
+project_dir = config.get("project_dir", config.DEFAULTS["project_dir"])
+active_project = config.get("active_project", "streetscapes")
+dbpath = Path(f"{project_dir}/{active_project}.duckdb")
```
BSchilperoort: Isn't the entire point of the config class to already handle the defaults?

E.g. in config.py it calls `cfg.setdefault(k, v)` after loading from file.

I think anywhere the code needs config settings, we should only use `config.get("<key>")`.

```suggestion
dbpath = Path(
    f"{config.get("project_dir")}/projects/{config.get("active_project")}.duckdb"
)
```

src/streetscapes/config.py
```diff
@@ -34,5 +34,6 @@ def get(key: str, default=None):
 
 def set(key: str, value):
     cfg = load()
-    cfg[key] = value
-    OmegaConf.save(cfg, CONFIG_FILE)
+    if key in DEFAULTS:
+        cfg[key] = value
+        OmegaConf.save(cfg, CONFIG_FILE)
```
BSchilperoort: This needs an `else` clause;


```suggestion
        OmegaConf.save(cfg, CONFIG_FILE)
    else:
        msg = f"Streetscapes has no config entry named '{key}'"
        raise KeyError(msg)
```

src/streetscapes/project.py
```diff
@@ -88,29 +88,28 @@ class Project:
...
-        self.root_dir = utils.ensure_dir(
+        # Directory for projects (databases + segmentations)
+        self.project_dir = Path(
             config.get(
-                "root_dir",
-                root_dir or pdirs.user_data_path("streetscapes"),
+                "project_dir",
+                utils.ensure_dir(
+                    project_dir or pdirs.user_data_path("streetscapes") / "projects"
+                ),
```
BSchilperoort: Why override normal default/fallback value. `project_dir` will always be in the config, so the default value here will do nothing.

BSchilperoort: `ensure_dir` should be called _after_ retrieving the `project_dir`.

src/streetscapes/project.py
```diff
@@ -88,29 +88,28 @@ class Project:
...
-            config.get("project_home", utils.ensure_dir(self.root_dir / "projects"))
-        )
 
         # Directory for cached data (images)
-        self.data_home = Path(
+        self.image_dir = Path(
             config.get(
-                "data_home",
+                "image_dir",
                 utils.ensure_dir(
```
BSchilperoort: Same problem here as with project_dir

BSchilperoort (changes requested):
We should also add the mapillary token to the config (and have a single source of truth).

Otherwise, see my comments. If these + the token are resolved we can merge this PR.

src/streetscapes/config.py
BSchilperoort: Yeah either add them to defaults, or make a list that is `DEFAULTS.keys() + ["other", "valid", "keys"]`.

BSchilperoort:
> add the mapillary token to the config

Maybe default value is grabbed from the env variables (and then will also be displayed to the user). With a fallback of an empty string.

Then in the mapillary download we just check if `if not config.get("mapillary_token"): raise Error` (empty strings resolve to false)

src/streetscapes/config.py
cantordust: OK, I added them to `DEFAULTS` - that's what I meant when I said above that every allowed key should have a default.🙂

src/streetscapes/config.py
```diff
@@ -23,16 +25,24 @@ def load() -> dict:
...
+    # Optional: remove deprecated keys that are not in DEFAULTS
+    deprecated_keys = set(cfg.keys()).difference(set(DEFAULTS.keys()))
+    for k in deprecated_keys:
+        cfg.pop(k)
+
     return cfg
 
 
-def get(key: str, default=None):
+def getopt(key: str, default=None):
```
BSchilperoort: Why rename these? I think this makes it less clear

BSchilperoort (approved):

PR #129, opened 2026-02-27
Register missing images when segmenting via the CLI

### LINKED ISSUE #127: Enable users to import local images into the streetscapes database format
Add a subcommand and appropriate options to allow users to add local images (not associated with a predefined online source, such as Mapillary) to a project, making them available for segmentation.

<hr>

Edit from @BSchilperoort :

Requirements;

- [x] add images through cli
- [x] have new "local" image source table that mostly mirrors exif (just like mapillary)

In addition, documentation is required, which

- [ ] guides users through using this functionality (maybe with wikimedia examples, such as [this one](https://commons.wikimedia.org/wiki/File:Arnstorf_-_Schloss_Mariakirchen_westlich,_Aussenansicht.JPG), or [this one](https://commons.wikimedia.org/wiki/File:Bad_S%C3%A4ckingen_%E2%80%94_Rathausplatz_3_und_M%C3%BCnsterplatz_34.JPG))
- [ ] points users to the right place for manually adding EXIF data, or a simple example for one file with e.g. Pillow.

cantordust:
Adds a registration step to register all missing images into the database when segmenting one or more images via the CLI.


BSchilperoort:
Can you expand on when/why this might be necessary?

cantordust:
The idea was to allow images to be registered while segmenting. The segmentation CLI takes a path, and we cannot guarantee that it would be a path that the project structure knows about. So you could point it to a directory containing some manually downloaded images, in which case it would be good to register the images with the active project before segmenting, otherwise we would lose track of them.

That said, the `register` argument defaults to `False` but perhaps it should be `True`🤔.

BSchilperoort:
> The segmentation CLI takes a path, and we cannot guarantee that it would be a path that the project structure knows about. So you could point it to a directory containing some manually downloaded images, in which case it would be good to register the images with the active project before segmenting, otherwise we would lose track of them.

But would it then not make more sense to remove the `path` argument? So users can only segment images that are already registered in the project? Because this way there are two ways to do the same thing (add images manually, add them via segmentation). IMO it would be better if there only was one way to register images, for simplicity sake and to avoid weird edge cases/bugs.

Seeing as the segmentation results are only stored in the database as polygons, and no where else, I don't think segmenting images that are not in the database makes sense anymore.

cantordust:
I'm OK with that, and removing the `path` argument would indeed make things easier. 

I can close this PR and make a new one, or just update this one - let me know which you prefer.

BSchilperoort:
> I'm OK with that, and removing the `path` argument would indeed make things easier.

OK great thanks!


> I can close this PR and make a new one, or just update this one - let me know which you prefer.

A new PR is probably cleaner

PR #131, opened 2026-03-03
Faster cli

### LINKED ISSUE #94: Improve CLI responsiveness
The CLI feels quite sluggish, as every command takes a while to load, even `streetscapes --help` or `streetscapes config list`.

One of the reason for this is that a lot of big packages are imported when the CLI is first built by click (main culprit being `torch`...).

Just delaying the torch import in `utils/functions.py` improves the `streetscapes config list` from 2.4s to 1.0s.

BSchilperoort:
Cleaner implementation than #124 . Closes #94 

I think all commands except `streetscapes segment-images` are now fast. I tried to mess with the code a minimal amount.

### DIFF
--- a/src/streetscapes/cli/database.py
+++ b/src/streetscapes/cli/database.py
@@ -18,7 +18,7 @@ def _get_db() -> "ibis.BaseBackend":
     import ibis
 
     dbpath = Path(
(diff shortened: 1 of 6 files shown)


BSchilperoort:
I tested the main functionalities, everything runs fine. I'll just merge and if any issues arise we'll fix those.

PR #135, opened 2026-03-20
Changed the default run name to model + timestamp

### LINKED ISSUE #103: Generate a human-readable run name for front end
Currently we refer to segmentation runs with a UUID. In the demo on 17 february we were asked if this could be a human-readable name instead (e.g. `modelname-timestamp` by default).

In my view we can;

- Add a "name" entry in the runs table
- Make this one the model name + time stamp by default, unless the user specifies a different run name
- Keep the uuid for internal/database use

cantordust:
Makes the default run name a bit more user-friendly.

### DIFF
--- a/src/streetscapes/models/bfms/cli.py
+++ b/src/streetscapes/models/bfms/cli.py
@@ -38,7 +38,8 @@ def cli(
     model = "bfms"
     model_params = {"model_id": model_id}
 
-    proj.add_run(run, model, model_params, overwrite)
+    result = proj.add_run(run, model, model_params, overwrite)
+    run = result.get("run")[0]
 
     if image_path is not None:
         image_paths = utils.get_image_paths(image_path)
--- a/src/streetscapes/models/bfms/cli.py
+++ b/src/streetscapes/models/bfms/cli.py
@@ -34,8 +34,6 @@ def cli(
     # ==================================================
     model = "bfms"
     model_params = {}
-    if run is None:
-        run = utils.uuid7(as_str=True)
 
     proj.add_run(run, model, model_params, overwrite)
 
--- a/src/streetscapes/project.py
+++ b/src/streetscapes/project.py
@@ -104,7 +104,7 @@ class Project:
 
         # Internal attributes
         # ==================================================
-        self._timestamp_resolution = "microseconds"
+        self._timestamp_resolution = "milliseconds"
 
         # Database connection
         self._con = ibis.duckdb.connect(
@@ -349,12 +349,14 @@ class Project:
         Returns:
             The data added to the database.
         """
+        ts = utils.iso_timestamp(self._timestamp_resolution)
+
         if run is None:
-            run = utils.uuid7(True)
+            run = f"{model or 'unknown'}@{ts.replace(' ', '_')}"
 
         data = {
             "run": [run],
-            "timestamp": [utils.iso_timestamp(self._timestamp_resolution)],
+            "timestamp": [ts],
             "model": [model],
             "metadata": [oj.dumps(metadata)],
         }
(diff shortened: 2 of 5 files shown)


src/streetscapes/models/bfms/cli.py
```diff
@@ -37,8 +37,6 @@ def cli(
 
     model = "bfms"
     model_params = {"model_id": model_id}
-    if run is None:
```
BSchilperoort, line 40: The `run` variable still isn't set here. So storing segmentation fails;
```
Constraint violation on 'segmentations': Constraint Error: NOT NULL constraint failed: segmentations.run
```

src/streetscapes/project.py
```diff
@@ -378,12 +380,13 @@ def add_runs(
         Returns:
             The data added to the database.
         """
-
+        ts = utils.iso_timestamp(self._timestamp_resolution)
```
BSchilperoort: I think having second precision would be better here.
As well as the standard separator "T".

The standard separator is actually being overwritten in the utils function;
```py
        ts = ts.isoformat(sep=" ", timespec=precision or "seconds")
```
sep defaults to "T" here.

I.e. just 
```py
datetime.now().isoformat(timespec="seconds")
```
suffices here I think.

cantordust: > The standard separator is actually being overwritten in the utils function;

I was toying with the idea of using the timestamp to create a hierarchy of directories for images in the format `date/time`, so that would have been convenient.

src/streetscapes/project.py
```diff
@@ -104,7 +104,7 @@ def __init__(
 
         # Internal attributes
         # ==================================================
-        self._timestamp_resolution = "microseconds"
+        self._timestamp_resolution = "milliseconds"
```
BSchilperoort, line 107: You changed the default, but still override it here with miliseconds, so the run name still becomes e.g. `bfms-2026-03-23T09:56:18.251`
```suggestion
```

src/streetscapes/utils/functions.py
```diff
@@ -30,6 +30,7 @@
 def iso_timestamp(
     precision: str = "seconds",
     fmt: str | None = None,
+    sep: str = "T",
```
BSchilperoort, line 33:
```suggestion
    sep: str = "T",
    utc: bool = True,
```

I think it would be better to store the run names as local machine time as that is more informative to the end user.

src/streetscapes/utils/functions.py
```diff
@@ -44,17 +45,19 @@ def iso_timestamp(
...
     """
 
-    ts = datetime.now()
+    ts = datetime.now(UTC)
 
     if fmt is not None:
         ts = datetime.strftime(ts, fmt)
     else:
-        ts = ts.isoformat(sep=" ", timespec=precision or "seconds")
+        ts = ts.isoformat(sep=sep, timespec=precision)[:-6]
```
BSchilperoort:
```suggestion
        ts = ts.isoformat(sep=sep, timespec=precision)
        ts = ts.split("+")[0]  # remove timezone info
```
more robust way to remove timestap (no magic numbers)

src/streetscapes/project.py
BSchilperoort:
```suggestion
        ts = utils.iso_timestamp(utc=False)
```

src/streetscapes/utils/functions.py
```diff
@@ -44,17 +45,19 @@ def iso_timestamp(
...
         precision: Precision for the timespec parameter.
         fmt: Explicit format.
+        sep: A custom separator for the default ISO format.
 
     Returns:
         The formatted timestamp.
     """
 
-    ts = datetime.now()
+    ts = datetime.now(UTC)
```
BSchilperoort:
```suggestion
    ts = datetime.now(UTC) if utc else datetime.now()
```

BSchilperoort (changes requested):
Some comments; 
- run name timestamp still shows miliseconds
- I think presenting the users with the local timestamp in the model run name is more useful (for other purposes it's better to store time in the database as UTC)

PR #137, opened 2026-03-20
Added a `rating` column to the `segmentations` table

### LINKED ISSUE #104: Add segmentation-level rating
At the demo on 17 february, we were asked to add a rating for segmentations.

This would mean adding a "rating" column to the segmentation table, of integer type (0-5).

cantordust:
Note: inserting segmentations into an existing database will fail. 

***Only for testing purposes***: The easiest way to test this would be to drop the `segmentations` table and run one of the segmentation models from the CLI. The table will be regenerated automatically with the new schema.


BSchilperoort (approved):
lgtm

BSchilperoort:
Let's merge #135 before this one

PR #138, opened 2026-03-23
List projects using CLI

### LINKED ISSUE #136: add `streetscapes database projects` command
Users should be able to see which projects exist.

`streetscapes database projects`

could then show a list of the projects and the number of images in each project

BSchilperoort:
Closes #136

```
$ streetscapes database projects
Projects stored in directory '/home/x/.local/share/streetscapes':

Name             | Images | Segmentation runs
-----------------+--------+------------------
proj-towns       |       0|                 0
proj-towns-small |    1013|                14
```

### DIFF
--- a/src/streetscapes/cli/database.py
+++ b/src/streetscapes/cli/database.py
@@ -1,6 +1,6 @@
 """CLI commands to view/manipulate the database."""
 from pathlib import Path
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Optional
 
 from cyclopts import App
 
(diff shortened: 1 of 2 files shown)


src/streetscapes/cli/database.py
```diff
@@ -71,3 +72,33 @@ def delete_segmentations(
...
+    if len(proj_dbs) > 0:
+        projects = {proj.stem: _get_table_counts(_get_db(proj.stem)) for proj in proj_dbs}
+
+        name_len = max([len(proj) for proj in projects]) + 1
+        print(
+            f"Projects stored in directory '{CFG.project_dir}':"
+        )
+        print()
+        print(f"{"Name".ljust(name_len)}| Images | Segmentation runs")
+        print("-"*name_len + "+" + "-"*8+ "+" + "-"*18)
```
cantordust: Why not use Rich here? Does it make the CLI too slow?

cantordust (approved):
Nice, works as intended!

src/streetscapes/cli/database.py
BSchilperoort: Could have used rich I guess. Was straightforward to code this though.

I don't think rich has a big overhead.

cantordust: Just a suggestion, please feel free to ignore 🙂.

```suggestion
        con = Console()
        tbl = Table("Name", "Images", "Segmentation runs")
        for proj, counts in projects.items():
            tbl.add_row(proj, counts[0], counts[1])
        con.print(tbl)
```

src/streetscapes/cli/database.py
```diff
@@ -1,6 +1,6 @@
 """CLI commands to view/manipulate the database."""
 from pathlib import Path
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Optional
 
 from cyclopts import App
```
cantordust, line 5: For the Rich version below.

```suggestion
from cyclopts import App

from rich.table import Table
from rich.console import Console
```

PR #139, opened 2026-03-23
Don't store masks as numpy arrays, only polygons

### LINKED ISSUE #125: Stop saving masks to disk
(Once the open PRs are merged) all segmentations are stored as polygons in the duckdb database. However, the masks are also still saved to disk in a compressed numpy array format. 

@Peter9192 Do we need to keep these? Or should we just stop storing them. They are not used anywhere else anymore.

BSchilperoort:
Closes #125


cantordust (approved):
Probably my first review of a PR with 0 additions.🤭

PR #140, opened 2026-03-25
Default image tags extracted from the source metadata

### LINKED ISSUE #109: Populate tags with default values
We can already set tags based on the metadata retrieved from mapillary, e.g.;

- "mapillary"
- "panoramic"
- "perspective", "fisheye", "equirectangular"/"spherical"

Maybe even "night" "morning" "day" and "evening" based on the captured_at timestamp 🤔 

These tags can then be used to filter data for segmenting / data export / explorer frontend.

cantordust:
Adds default tags for images downloaded from Mapillary. 

TODO: The same for other sources.

### DIFF
--- a/src/streetscapes/cli/download_images.py
+++ b/src/streetscapes/cli/download_images.py
@@ -59,9 +59,19 @@ def mapillary(
     image_data = []
     downloaded = 0
 
-    for idx, (uid, image_id, url, shard, location) in track(
+    for idx, rec in track(
         enumerate(records), "Downloading images...", total=len(records)
     ):
+
+        (
+            uid,
+            image_id,
+            url,
+            shard,
+            location,
+            is_pano,
+            camera_type,
+        ) = rec
 
         # Determine the shard
         output_dir = Path(image_dir)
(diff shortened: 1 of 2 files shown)


src/streetscapes/cli/download_images.py
```diff
@@ -73,15 +83,20 @@ def mapillary(
 
         # Download image
         meta = mapillary.download_image(url, output_dir, uid, skip_existing)
-
         meta.shard = str(shard)
+        tags = ['mapillary']
+        if is_pano:
+            tags.append('panoramic')
+        if camera_type is not None:
+            tags.append('camtype')
```
BSchilperoort:
```suggestion
            tags.append(camera_type)
```

BSchilperoort (approved):
Found one small typo ;)

Also, I notice you using single quotes ' while the ruff formatter will change those to "

src/streetscapes/cli/download_images.py
cantordust: Yeah, it was supposed to be f"{camtype}", and then I renamed the variable 😅

PR #141, opened 2026-03-25
Backend fixes

### LINKED ISSUE #142: Image cache is not used when duckdb database is removed
To reproduce;

- have existing project where you previously downloaded images
- rm project-name.duckdb
- create new project
- fetch metadata for same bbox as before
- download images

All the images will be downloaded from mapillary again.

The problem here seems to be circular reasoning;
- you need the image UUID to find it in the cache, as the uuid is its name
- to generate the image UUID you need to have access to the file
- to find the file you need the UUID
- etc.

Thus when removing the database, or switching to a different active project, it is impossible to reconstruct the database using the cache, and it will always be re-downloaded.

### LINKED ISSUE #143: Change repo license from CC-BY to Apache 2.0
At the eScience Center we default to Apache 2.0 as license.

BSchilperoort:
Todo;

- [x] stats for segmentations
- [x] filtering for segmentations
- [x] allow setting image rating
- [x] allow settings image tags
- [x] allow setting segmentation rating

The backend can now serve data and filter the requested images. More filters have been implemented already as well.

Setting image ratings/tags/notes is also enabled now.


src/streetscapes/explorer/server.py
sverhoeven: https does not work change to http

src/streetscapes/explorer/data.py
```diff
@@ -13,13 +13,19 @@ class Bbox(BaseModel):
...
+    # image level filters
+    image_ratings: list[int] = Field(default=[])
     sources: list[str] = Field(default=[])
     tags: list[str] = Field(default=[])
-    compass_angle: list[float] = Field(default=[0, 360])
+    # metadata level filters
+    compass_angle: list[float] = Field(default=[0, 360])  # TODO: add to backend filter
     date_range: tuple[datetime, datetime] = Field(default=(datetime(1826,1,1),datetime.now()))
-    panoramic: list[int] = Field(default=[])
+    panoramic: list[int] = Field(default=[])  # TODO: add to backend filter
```
sverhoeven: Can be done with tags, remove this prop

src/streetscapes/explorer/server.py
```diff
@@ -130,50 +135,141 @@ def _get_metadata(id: str) -> ImageMetadata:
     uuid = UUID(id)
     source = _get_source(uuid)
 
+    con = _open_db(dbpath)
     match = con.table(source).image == uuid
-    data = con.table(source).filter(match).to_pandas()
+    metadata = con.table(source).filter(match).to_pandas()
+    match = con.table("images").uuid == uuid
```
sverhoeven: Move this up and remove _get_source() as imgdata contains source column

src/streetscapes/explorer/server.py
```diff
@@ -130,50 +135,141 @@ def _get_metadata(id: str) -> ImageMetadata:
...
+    ):
+        segmentations = con.table("segmentations")
+        # optionally filter for models
+        if len(filter.models) > 0:
+            runs = con.table("runs")
+            runs = runs.filter(runs.model.isin(filter.models))
+            segmentations = segmentations.filter(segmentations.run.isin(runs.run))
+
+        for label in filter.labels:
+            print(f"filtering for label {label}")
```
sverhoeven:
```suggestion
```

src/streetscapes/explorer/server.py
```diff
@@ -130,50 +135,141 @@ def _get_metadata(id: str) -> ImageMetadata:
...
+        if len(filter.model_runs) > 0:
+            segmentations = segmentations.filter(
+                segmentations.run.isin(filter.model_runs)
+            )
+        if len(filter.segmentation_ratings) > 0:
+            segmentations = segmentations.filter(
+                segmentations.run.isin(filter.segmentation_ratings)
+            )
+
+        valid_images = set(segmentations.image.to_pandas())
```
sverhoeven: This moves all rows from db to py, can you do `SELECT DISTINCT id`?

sverhoeven: Or just add `.select(imgid)`

src/streetscapes/explorer/server.py
```diff
@@ -197,35 +317,23 @@ async def project():
...
-            "terrain",
-            "pedestrian-area",
-        ],
-        model_run_names=["manual"],
-        image_sources=[
-            "mapillary",
-        ],
-        date_range=(datetime(2026, 1, 19), datetime(2026, 1, 20)),
-        models=["DinoSAM", "maskformer", "bfms", "manual"],
+        tags=_get_unique_tags(con),
```
sverhoeven: Use distinct in the ibis https://ibis-project.org/reference/expression-tables.html#ibis.expr.types.relations.Table.distinct and select(...)

sverhoeven (changes requested):

src/streetscapes/explorer/server.py
BSchilperoort: Modified this to
```py
valid_images = segmentations.distinct(on="image").image
```

src/streetscapes/explorer/server.py
BSchilperoort: distinct only works on the whole column value, not the individual elements in the array of each row.

To do this operation with ibis, we need to first call `collect`, then flatten this, and then get the unique values. Turns out this takes longer than just using pandas;

image

The `to_pandas` call is always needed to get the list of strings.

BSchilperoort: Pandas native alternative for getting the unique labels;

image

BSchilperoort:
@cantordust I decided to also fix #142 in [af98d75](https://github.com/Urban-M4/streetscapes/pull/141/commits/af98d7556870b4058b0e45dfb9678615216011ce). Besides just an image `{uuid}.jpg` being written to file, I also write the Mapillary image ID to file, with the UUID as content. So when clearing the database, images do not need to be redownloaded anymore.

src/streetscapes/cli/download_images.py
```diff
@@ -9,12 +9,47 @@
...
 logger = logging.getLogger(__name__)
 
 download_images_cli = App(help="Download images from various sources.")
 
 
+def _validate_uuid(uid: str, output_dir: Path) -> bool:
+    img_file = (output_dir / uid).with_suffix(".jpg")
+    if img_file.exists() and uid == utils.get_image_uuid(img_file):
+        return True
+    return False
```
cantordust:
```suggestion
    return img_file.exists() and uid == utils.get_image_uuid(img_file)
```

pyproject.toml
sverhoeven:
```suggestion
license = { file = "LICENCE" }
```

src/streetscapes/explorer/data.py
```diff
@@ -13,13 +13,17 @@ class Bbox(BaseModel):
...
+    # image level filters
+    image_ratings: list[int] = Field(default=[])
     sources: list[str] = Field(default=[])
     tags: list[str] = Field(default=[])
-    compass_angle: list[float] = Field(default=[0, 360])
+    # metadata level filters
     date_range: tuple[datetime, datetime] = Field(default=(datetime(1826,1,1),datetime.now()))
-    panoramic: list[int] = Field(default=[])
+    # segmentation level filters
+    models: list[str] = Field(default=[])
```
sverhoeven:
```suggestion
    models: list[str] = []
```

See https://docs.pydantic.dev/latest/concepts/fields/#mutable-default-values

src/streetscapes/explorer/data.py
```diff
@@ -13,13 +13,17 @@ class Bbox(BaseModel):
 
 
 class FilterParams(Bbox):
-    ratings: list[int] = Field(default=[])
-    model_runs: list[str] = Field(default=[])
+    # image level filters
+    image_ratings: list[int] = Field(default=[])
```
sverhoeven: create Rating type as rating must 0..5

Use https://github.com/annotated-types/annotated-types?tab=readme-ov-file#gt-ge-lt-le

sverhoeven (approved):
Looks good.
Besides license, just a couple of nitpicks.

PR #145, opened 2026-04-01
Fix typing

### LINKED ISSUE #144: Add type checker to CI
We add typing to functions to help with writing code and debugging. However, these types are not always correct. 

We should setup `mypy` and have it validate typed functions. We would not want to enforce typing everywhere as that's quite impossible with the databases.

BSchilperoort:
Merge after #141 

Closes #144


sverhoeven (commented):
Please, make use of either `Optional` or `| None` now both are used. Even on same line sometimes.

src/streetscapes/cli/config.py
```diff
@@ -24,13 +24,13 @@ def set_config(key: str, value: str):
     from streetscapes.project import Project
 
     if not hasattr(CFG, key):
-        print(f"No config value for '{key}'", err=True)
-        raise SystemExit(code=1)
+        print(f"No config value for '{key}'")
+        raise SystemExit
```
sverhoeven:
```suggestion
        sys.exit(1)
```

BSchilperoort:
> Please, make use of either `Optional` or `| None` now both are used. Even on same line sometimes.

There seemed to be more ` | None` occurrences than `Optional[` so I decided to switch everything to `| None`

PR #150, opened 2026-04-10
Import local images via the CLI

cantordust:
Added a subcommand allowing users to import local images (i.e., ones that are not associated with a predefined online source, such as Mapillary).

The `local` table is populated automatically using EXIF metadata (including GPS coordinates).


src/streetscapes/cli/images.py
```diff
@@ -0,0 +1,30 @@
...
+from streetscapes import CFG
+
+if TYPE_CHECKING:
+    import ibis
+
+image_cli = App(help="Perform various operations on local collections of images.")
+
+
+@image_cli.command(name="add")
+def add_images(
```
BSchilperoort, line 10: Would probably be nicer to merge this into a single command instead of a separate cli section. Especially as we don't have nice completion/suggestions.

```
streetscapes images add
```
is just a bit more hassle than
```
streetscapes add-images
```

because first time users need to go through
```
streetscapes --help
streetscapes images --help
streetscapes images add --help
```
before knowing how to structure the command.

src/streetscapes/utils/functions.py
```diff
@@ -612,3 +615,74 @@ def uuid7(as_str: bool = False) -> uuid.UUID | str:
...
+    # Extract the tags that we are interested in
+    lon = tags.get("GPS GPSLongitude")
+    lon_ref = tags.get("GPS GPSLongitudeRef")
+    sign = -1 if lon_ref == "W" else 1
+    lon = 0.0 if lon is None else to_deg(lon.values, sign)
+
+    lat = tags.get("GPS GPSLatitude")
+    lat_ref = tags.get("GPS GPSLatitudeRef")
+    sign = -1 if lat_ref == "S" else 1
+    lat = 0.0 if lat is None else to_deg(lat.values, sign)
```
BSchilperoort: Would be nice to add this logic to the `to_deg` function, i.e.

```py
def to_deg(dms: list[int | float], reference: Literal["N","E","S","W"]) -> float:
    """
    Convert [deg, min, s] to degrees.

    Args:
        dms: A list containing degrees, minutes and seconds.
        reference: Coordinate reference (N, E, S, or W).

    Returns:
        Latitude or longitude coordinate in decimal degrees.
    """
    sign = -1 if reference in ("W", "S") else 1
    deg = dms[0] + dms[1] / 60 + float(dms[2]) / 3600
    return sign * deg

# So we can do;
lat = to_deg(
    tags.get("GPS GPSLatitude").values,
    tags.get("GPS GPSLatitudeRef")
)
lon = to_deg(
    tags.get("GPS GPSLongitude").values,
    tags.get("GPS GPSLongitudeRef")
)
```

src/streetscapes/utils/functions.py
```diff
@@ -612,3 +615,74 @@ def uuid7(as_str: bool = False) -> uuid.UUID | str:
...
+
+def to_deg(dms: list[int | float], sign: int = 1) -> float:
+    """
+    Convert [deg, min, s] to degrees.
+
+    Args:
+        dms: A list containing degrees, minutes and seconds.
+
+    Returns:
+        A single float.
```
BSchilperoort:
```suggestion
        Latitude or longitude coordinate in decimal degrees.
```

src/streetscapes/utils/functions.py
BSchilperoort, line 1: Would be nicer to move the new functions to
`src/streetscapes/utils/exif-parsing.py`
so the utils are organized by functionality

src/streetscapes/project.py
```diff
@@ -116,6 +121,25 @@ class Project:
             },
             "init": [],
         },
+        "local": {
+            "schema": {
+                "image": "UUID PRIMARY KEY",
+                "make": "STRING",
+                "model": "STRING",
+                "orientation": "UBIGINT",
```
BSchilperoort, line 129: This is the image orientation, right? Would be nicer to automatically rotate it correctly... 
Maybe just make an issue out of this as a future enhancement

src/streetscapes/project.py
```diff
@@ -116,6 +121,25 @@ class Project:
...
+                "timestamp": "TIMESTAMP",
```
BSchilperoort, line 130:
```suggestion
                "captured_at": "TIMESTAMP",
```
Would be nicer to store this under the same name as the Mapillary table.

Although the mapillary time value is stored as an int...

src/streetscapes/project.py
```diff
@@ -580,30 +609,39 @@ def get_segmentation_status(
...
+        exif_data: list[dict] | None = None,
         overwrite: bool = False,
     ):
         """
         Register downloaded (local) images into the database.
 
         Args:
-            data: A dictionary of image information.
+            images: A list of dictionaries containing image information.
+            exif_data: Metadata extracted from the images' EXIF tags.
```
BSchilperoort:
```suggestion
            exif_data: Metadata extracted from the images' EXIF tags. Note: only used for locally imported images.
```

BSchilperoort (commented):
Some first comments, still need to run the code to test.

src/streetscapes/cli/images.py
cantordust: Sure, I don't feel strongly either way, so I'd be happy to turn it into a single command.

src/streetscapes/utils/functions.py
cantordust: In principle I completely agree, but I think that we can do that in one go in a separate issue since this is not the only function that needs to be put into a separate module.

src/streetscapes/utils/functions.py
cantordust: Makes sense, I'll add that!

src/streetscapes/cli/images.py
cantordust: On second thought, I'm keen to stick with this design. It makes sense to me to group actions under `images` - just makes it much more future-proof. Otherwise we'll end up with a flat list of gajillion subcommands (`add-images`, `delete-images`, etc.). What do you think? Apropos, the `database` CLI follows the same pattern.

src/streetscapes/project.py
cantordust: Yeah, that's why I renamed it - it's not the same value any more.

cantordust:
Ready for another review!

- I kept the CLI format for adding images as is, see the justification in the comment above. But I'd be happy to discuss further!🙂
- I haven't restored the `captured_at` database column because it's effectively a different field now. Again, happy to discuss and revise.😊

src/streetscapes/cli/images.py
BSchilperoort: Guess it's fine for now

src/streetscapes/cli/images.py
```diff
@@ -0,0 +1,20 @@
...
+    path: Path,
+    project: str | None = None,
+    shard: str | None = None,
+    overwrite: bool = False,
+):
+    """Add images from a local directory."""
```
BSchilperoort: Docs for the arguments are needed here. Currently the CLI has no documentation at all;

```
$ streetscapes images add --help
Usage: streetscapes images add PATH [ARGS]

Add images from a local directory.

╭─ Parameters ───────────────────────────────────────────────────────────────────╮
│ *  PATH --path            [required]                                           │
│    PROJECT --project                                                           │
│    SHARD --shard                                                               │
│    OVERWRITE --overwrite  [default: False]                                     │
│      --no-overwrite                                                            │
╰────────────────────────────────────────────────────────────────────────────────╯
```

maybe also add some explanation for the use cases.

src/streetscapes/project.py
```diff
@@ -593,26 +617,37 @@ def get_segmentation_status(
...
-        data: dict[str, list] = {column: [] for column in self.schema("images")}
+        # Entries for the `images` table.
+        img_data: dict[str, list] = {column: [] for column in self.schema("images")}
 
         for image in images:
-            for column in data:
-                data[column].append(image.get(column))
+            for column in img_data:
+                img_data[column].append(image.get(column))
+        self.update_table("images", img_data, overwrite)
```
BSchilperoort, line 638: Would be nice to at least use `is_pano` to add the tag `panoramic` to an image. Currently no tags are automatically populated.

BSchilperoort: Hm, there is an `is_pano` entry in the local images database, but this does not seem to be set anywhere. I looked around and while there is no true standard, the most commonly supported convention is this; https://developers.google.com/streetview/spherical-metadata with XMP (not exif).

We should adopt that so users can properly add panoramic images. You can make a new issue for this.

BSchilperoort (approved):
Tried it. Seems to work. 

Can you add documentation? Adding the command to the README is OK, and add some more info to the `streetscapes images add` cli.

Currently you set `is_pano` to always be false with user added images. However, there is a semi-standard convention that seems to be widely adopted. It would be nice if this was properly loaded from the images too. Can be in new PR.

Feel free to merge after improving docs and opening issue on pano xmp data.

src/streetscapes/project.py
cantordust: I did add it as part of the PR about default tags:

https://github.com/Urban-M4/streetscapes/tree/109-default-tags

But this is from metadata provided by Mapillary. A separate issue is probably warranted.

PR #155, opened 2026-04-30
Fix intake mapillary

BSchilperoort:
@cantordust had to make this fixes to be able to download a larger dataset over Vienna.

### DIFF
--- a/src/streetscapes/project.py
+++ b/src/streetscapes/project.py
@@ -664,6 +664,15 @@ class Project:
 
         df.insert(loc=0, column="uuid", value=None)
 
+        expected_colcount = len(self.core_tables["mapillary"]["schema"])
+        if df.columns.size != expected_colcount:
+            msg = (
+                "Missing columns in image. Skipping..."
+                f"Image is available at {df.get("thumb_2048_url")}"
+            )
+            logger.error(msg)
+            return None
+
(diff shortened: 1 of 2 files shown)


src/streetscapes/sources/mapillary.py
```diff
@@ -203,7 +203,10 @@ def unpack_geometry(geometry):
             return None
 
         df["geometry"] = df["geometry"].apply(unpack_geometry)
-        df["computed_geometry"] = df["computed_geometry"].apply(unpack_geometry)
+        if df.get("computed_geometry") is None:
+            df["computed_geometry"] = None
```
BSchilperoort, line 207: Sometimes computed_geometry was missing during a bigger download

src/streetscapes/project.py
```diff
@@ -664,6 +664,15 @@ def ingest_mapillary(self, df: DataFrame, table: str = "mapillary"):
 
         df.insert(loc=0, column="uuid", value=None)
 
+        expected_colcount = len(self.core_tables["mapillary"]["schema"])
+        if df.columns.size != expected_colcount:
```
BSchilperoort, line 668: Some images consistently return too little metadata. I decided to just skip them to make the download not fail

cantordust (approved):

PR #161, opened 2026-05-18
Serve images to front end through API

### LINKED ISSUE #154: Pass images to front-end through api
Mapillary changed its image URLs so older projects won't have correct urls anymore. This means the front-end cannot load the images. We do already download the images locally however, so it would be better to just use those instead.

BSchilperoort:
Closes #154

### DIFF
--- a/src/streetscapes/explorer/server.py
+++ b/src/streetscapes/explorer/server.py
@@ -14,7 +14,7 @@ import pandas as pd
 import uvicorn
 from brotli_asgi import BrotliMiddleware
 from cyclopts import App, Parameter
-from fastapi import FastAPI, HTTPException
+from fastapi import FastAPI, HTTPException, Response
 from fastapi.middleware.cors import CORSMiddleware
 from fastapi.params import Query
 from shapely.ops import transform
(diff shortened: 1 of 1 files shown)


sverhoeven (commented):
I was able to fetch image:

Image

Let me make a streetscapes-explorer pr that uses the new endpoint.

src/streetscapes/explorer/server.py
```diff
@@ -332,6 +361,16 @@ async def fetch_image_metadata(image_id: str) -> ImageMetadata:
...
 
 
+@app.get(
+    "/images/{image_id}/img",
+    responses = {200: {"content": {"image/jpeg": {}}}},
+    response_class=Response
+)
+async def fetch_image(image_id: str) -> Response:
+    """Get all metadata associated with a certain image, including segmentations."""
+    return Response(_get_image(image_id), media_type="image/jpeg")
```
sverhoeven: Use [FileResponse](https://fastapi.tiangolo.com/advanced/custom-response/#fileresponse), instead of Response, as it works async.

sverhoeven: Or use something similar to https://fastapi.tiangolo.com/tutorial/static-files/, it uses [notmodifiedresponse](https://github.com/Kludex/starlette/blob/b8fa5140d2ef9f22483d777e936ab4c2df897179/starlette/staticfiles.py#L181-L183) for better caching.

src/streetscapes/explorer/server.py
sverhoeven, line 354: It would be nice to include local url in response, same for fetch_image_metadata

sverhoeven (approved):
Works, but some inline suggestions which do not block.

webapp pr has been merged so please merge this one as well.

sverhoeven (approved):

PR #163, opened 2026-05-18
Homogenize local schema, support local images in front-end

### LINKED ISSUE #159: make the backend be able to serve from the "local" table (currently only supports mapillary).

### LINKED ISSUE #162: Homogenize different data sources
I am trying to make the backend work, but I have to add a lot of branches in the code to deal with the small differences between the mapillary and local source tables.

The following will be needed to make the local source look more like the mapillary table;

- [ ] Add `compass_angle` entry ("GPS GPSImgDirection" in the exif data).
- [x] Store timestamp as an int, under the name `captured_at` (in ms since epoch instead of datetime). Alternatively we convert the mapillary timestamp to a datetime instead.

this would greatly simplify the code needed to work with the two different data sources. For any future sources this subset of essential metadata should be populated as well, using the same names and data types.

BSchilperoort:
Closes #162, #159

### DIFF
--- a/src/streetscapes/project.py
+++ b/src/streetscapes/project.py
@@ -127,7 +127,8 @@ class Project:
                 "make": "STRING",
                 "model": "STRING",
                 "orientation": "UBIGINT",
-                "timestamp": "TIMESTAMP",
+                "captured_at": "UBIGINT",
+                "compass_angle": "FLOAT8",
                 "width": "UBIGINT",
                 "height": "UBIGINT",
                 "altitude": "FLOAT4",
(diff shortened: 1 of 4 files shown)


src/streetscapes/project.py
```diff
@@ -127,7 +127,8 @@ class Project:
                 "make": "STRING",
                 "model": "STRING",
                 "orientation": "UBIGINT",
-                "timestamp": "TIMESTAMP",
```
cantordust, line 130: Should we keep `timestamp` as well? `captured_at` is an integer that would be meaningless to most people.

cantordust (approved):
Looking good!🚀

src/streetscapes/project.py
BSchilperoort: IMO it would have been better to store the captured_at data as a timestamp only, everywhere 😅. Storing the same information in two formats seems silly. Maybe better as a new PR. I'll open an issue

BSchilperoort:
Thanks for the review!

PR #171, opened 2026-06-19
Reproducible image UUIDs

### LINKED ISSUE #170: Make image UUIDs reproducible
Open images as NumPy arrays before computing the hash and subsequently the UUID.

cantordust:
Fixed a bug where images with different EXIF data would produce different UUIDs.

Closes #170.

### DIFF
--- a/src/streetscapes/utils/functions.py
+++ b/src/streetscapes/utils/functions.py
@@ -9,6 +9,8 @@ from datetime import datetime, UTC
 from hashlib import sha256
 from pathlib import Path
 
+from PIL import Image
+import numpy as np
 import filetype as ft
 import seedir as sd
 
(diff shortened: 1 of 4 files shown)


src/streetscapes/utils/functions.py
```diff
@@ -474,13 +474,18 @@ def get_image_hash(image: str | Path | bytes) -> bytes:
...
 
-    if isinstance(image, str | Path):
-        image = Path(image).read_bytes()
-
     if not ft.is_image(image):
         raise ValueError("The provided file is not an image.")
 
-    return sha256(image).digest()
+    from PIL import Image
+    import numpy as np
```
BSchilperoort: These can be at top of file. Both aren't heavy imports

BSchilperoort (approved):
Tested it with old and new images from mapillary, that had differing UUIDs. If I now generate the uuids for the images they match. Thanks for fixing this!

PR #172, opened 2026-08-26
Fix mapillary pipeline

cantordust:
- Fix for tile size restriction in the Mapillary pipeline.
- Fix for the `skip_existing` switch.

### DIFF
--- a/src/streetscapes/cli/download_images.py
+++ b/src/streetscapes/cli/download_images.py
@@ -2,10 +2,9 @@ import logging
 import os
 from pathlib import Path
 
+import typer
 from cyclopts import App
 from rich.progress import track
-
-import typer
 
 from streetscapes import CFG, utils
 from streetscapes.cli.console import console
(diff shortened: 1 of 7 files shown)


src/streetscapes/cli/download_images.py
```diff
@@ -114,7 +113,14 @@ def mapillary(
...
 
         if not skip_existing or not _existing_img_valid(uid, image_id, output_dir, skip_existing):
-            uid = mapillary.download_image(url, output_dir, image_id, uid, skip_existing).uid
+            try:
+                img_meta = mapillary.download_image(
+                    url, output_dir, image_id, uid, skip_existing=skip_existing
+                )
+                uid = img_meta.uid
+            except Exception as e:
+                logger.error(e)
```
BSchilperoort, line 122: This logger output is not reaching the end user by default...

Adding an exception on line 120, and then testing downloading images makes the download fail silently.
The end user should be clearly notified when images aren't being downloaded.

BSchilperoort (commented):
Downloading seems to work fine on my machine with this PR.

~~I did identify an issue with the log output not showing up to stdout, so users aren't notified that e.g. downloads fail.~~

src/streetscapes/cli/download_images.py
BSchilperoort: nvm my error

BSchilperoort (approved):

Issue #106, opened 2026-02-19
Ease installation


BSchilperoort:
Currently to install the package, users first have to (in a clean environment):

- pip install tranformers
- pip install tokenizers
- pip install the package itself

This is not very user-friendly, and the dependency resolution takes a while.

We could solve this by;
- Pinning tranformers/tokenizers (solves installation difficulty)
- Make uv lock file and modify user instructions to refer to this (solves installation difficulty + installation time)

BSchilperoort:
Fixed in #173; install works fine in Python 3.14. Uv lock file is also present.

PR #173, opened 2026-08-26
Fix install (and move to Python 3.14)

BSchilperoort:
There have been some challenges installing the package due to some transformers/tokenizers dependency conflict nonsense.
I have tested installing the package in Python 3.14, which works flawlessly, making installation a lot easier.

Fiona had to be removed as dependency (unused and not 3.14 compatible...).

Moving to 3.14 should ensure longer term stability of the package. 

Solves #106 partially.


cantordust (approved):
Looks good, thank you!😊

PR #177, opened 2026-08-26
Please Ruff

### LINKED ISSUE #165: Ruff linter runs on CI but is allowed to fail
> `--exit-zero` will cause Ruff to exit with a status code of 0 even if violations were found.

Image

I don't think Ruff is very useful like this 🤔

BSchilperoort:
Linters are only useful if you look at their output. This PR enforces Ruff in CI and fixes all issues in the code that built up over time.


src/streetscapes/sources/mapillary.py
```diff
@@ -41,12 +46,17 @@ class MapillaryClient:
         # Or fetch directly as GeoDataFrame
         gdf = client.fetch_metadata_bbox_gpd(bbox)
 
-    Methods
+    Methods:
     -------
```
cantordust: I think the Google style doesn't have these underlines.

src/streetscapes/streetview/image_processing.py
cantordust:
```suggestion
    Args:
```

cantordust (changes requested):
I think Ruff missed some corner cases in the docstrings, like underlines (`-----`) and some classes where we have `Parameters` instead of `Args`, but otherwise it looks fine to me.

BSchilperoort:
> I think Ruff missed some corner cases in the docstrings, like underlines (`-----`) and some classes where we have `Parameters` instead of `Args`, but otherwise it looks fine to me.

Well spotted. I think I fixed them now.

cantordust (approved):
🚀

PR #179, opened 2026-08-27
SAM3 support

cantordust:
Added support for image segmentation with SAM3.

Issue #174, opened 2026-08-26
Prune dependencies


BSchilperoort:
Once the example notebooks have been refactored to the new format, we should go over all dependencies and check if they are still required. Many are outdated and not used anymore.

seemingly unused:
- black
- geodatasets
- h3
- humanize
- tabulate
- timezonefinder

BSchilperoort:
also consider switching to [dependency groups](https://packaging.python.org/en/latest/specifications/dependency-groups/) down the track.

PR #180, opened 2026-08-28
Use cloudpickle for Ray worker-node comms instead of orjson

BSchilperoort:
Review after #179 

In addition, I improved the experience for AMD GPU users.

Also, during testing I found a bug in DinoSAM (if no instances of the prompt are found, the lack of results wasn't handled properly). Fixed in this PR as well.

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -23,7 +23,7 @@ jobs:
           enable-cache: true
 
       - name: Install the project
-        run: uv sync --locked --extras sam3,dev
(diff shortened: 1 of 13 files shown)


cantordust (approved):
Thank you, looks good. Perhaps it would be good to switch to [dependency groups](https://packaging.python.org/en/latest/specifications/dependency-groups/) down the track.

BSchilperoort:
> . Perhaps it would be good to switch to [dependency groups](https://packaging.python.org/en/latest/specifications/dependency-groups/) down the track.

Yeah we could tackle that with #174

Issue #181, opened 2026-08-28
Sanitize mapillary metadata API output


BSchilperoort:
Mapillary data seems to not follow their schema very well. Fields can be wrong, or even have incorrect data in fields (e.g. the image URL in the ID field).

We should use a Pydantic model to validate the query results and use them only if they pass the validation. This way we can also skip single images within a tile.

PR #182, opened 2026-08-31
Various small fixes

BSchilperoort:
- mapillary fetch-metadata could fail. Patched now, proper fix is described in #181 
- SAM3 segmentation could fail if no instances were found matching the prompt
- `--prompt "red wall, blue sky"` would be converted to `["red", "wall", "blue", "sky"]` before being passed to a segmentation model. Will now return `["red wall", "blue sky"]` to allow for more complex instance detection in SAM3.

### DIFF
--- a/src/streetscapes/project.py
+++ b/src/streetscapes/project.py
@@ -688,13 +688,14 @@ class Project:
     def ingest_mapillary(self, df: DataFrame, table: str = "mapillary"):
         """Ingest a DataFrame of Mapillary metadata."""
         # Ensure that that the camera_parameters column is a list of floats
-        df["camera_parameters"] = df["camera_parameters"].apply(
-            lambda params: (
-                [float(params)]
-                if isinstance(params, int | float)
-                else list(map(float, params))
+        if df.get("camera_parameters") is not None:
+            df["camera_parameters"] = df["camera_parameters"].apply(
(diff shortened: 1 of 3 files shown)


src/streetscapes/project.py
```diff
@@ -711,7 +712,14 @@ def ingest_mapillary(self, df: DataFrame, table: str = "mapillary"):
 
         # TODO: consider configurable duplicate behaviour (REPLACE or IGNORE)
         # TODO do not depend of order in df columns, but used named columns in SQL query
-        self._con.raw_sql(f"INSERT OR IGNORE INTO {table} FROM metadata_tile")
+        try:
+            self._con.raw_sql(f"INSERT OR IGNORE INTO {table} FROM metadata_tile")
+        except Exception as err:
+            if "Conversion Error" in str(err):
```
cantordust, line 718: Is this just for fields that don't match their types as declared by the Mapillary schema?

cantordust (approved):
Thank you for adding these fixes!🎉

src/streetscapes/project.py
BSchilperoort: Yeah. Mapillary will raise some Conversion Error if the field doesn't match the type it should be (e.g. a field with string data where it should be an int)

PR #188, opened 2026-09-02
Fix local images in explorer, improve documentation

### LINKED ISSUE #127: Enable users to import local images into the streetscapes database format

BSchilperoort:
9034eda broke local images in explorer

Closes #127

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -124,7 +124,6 @@ dev = [
 ]
 sam3 = [
     "ultralytics>=8.3.237",
-    "clip@git+https://github.com/ultralytics/CLIP.git"
 ]
 
 [tool.uv.sources]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -122,7 +122,10 @@ dev = [
     "types-shapely",
     "types-requests",
 ]
-sam3 = ["ultralytics>=8.3.237"]
+sam3 = [
+    "ultralytics>=8.3.237",
+    "clip@git+https://github.com/ultralytics/CLIP.git"
+]
 
 [tool.uv.sources]
 torch = [{ index = "pytorch-rocm", extra = "rocm" }]
--- a/README.md
+++ b/README.md
@@ -17,12 +17,12 @@ Requirements: Python 3.14
 In a Python virtual environment, or conda environment do:
 
 ```bash
-pip install streetscapes[sam3]
+pip install streetscapes==1.0.0a0[sam3]
 ```
 
 To install for AMD GPUs (ROCm), do;
 ```bash
-uv pip install streetscapes --extra rocm,sam3
+uv pip install streetscapes==1.0.0a0 --extra rocm,sam3
 ```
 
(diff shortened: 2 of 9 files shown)


pyproject.toml
```diff
@@ -122,7 +122,10 @@ dev = [
     "types-shapely",
     "types-requests",
 ]
-sam3 = ["ultralytics>=8.3.237"]
+sam3 = [
+    "ultralytics>=8.3.237",
+    "clip@git+https://github.com/ultralytics/CLIP.git"
```
cantordust: This will most likely prevent the package from being published on PyPI:
- https://setuptools.pypa.io/en/latest/userguide/dependency_management.html#direct-url-dependencies
- https://peps.python.org/pep-0440/#direct-references

This is probably why the `ultralytics` package needs to install CLIP in some shifty way. I'm not sure why they can't just publish it to PyPI.

cantordust (changes requested):
Nice, mostly works as expected! Two comments:

- The second image failed to download multiple times (it created a zero-sized file). Eventually it worked (defying the old maxim that you shouldn't do the same thing multiple times and expect a different result), but I'm not sure why. Could it have something to do with the special characters in the URL?
- The CLIP dependency should probably be managed by the `ultralytics` package (see my comment).

pyproject.toml
BSchilperoort: Hm. That would be a problem.

> This is probably why the ultralytics package needs to install CLIP in some shifty way. 

Which doesn't seem to work reliably (e.g. on my laptop)...

BSchilperoort: Added a note to the readme instead

BSchilperoort:
> * The second image failed to download multiple times (it created a zero-sized file). Eventually it worked (defying the old maxim that you shouldn't do the same thing multiple times and expect a different result), but I'm not sure why. Could it have something to do with the special characters in the URL?

Could have to do with Wikimedia's api/rate limiting. I'll add links for users to download the images from the normal webpage as an alternative

cantordust (approved):
Looks good!🤓

PR #192, opened 2026-09-03
Remove Maskformer labels argument

### LINKED ISSUE #113: Maskformer ignores label argument passed through cli
If you run maskformer segmentation with, e.g., `streetscapes segment_images maskformer --labels building`, you _should_ get only the building labels, but instead this is ignored and the [full table of built-in labels](https://github.com/Urban-M4/streetscapes/blob/73beec98c6146ab493fdac04058d22941fd21652/src/streetscapes/models/maskformer/model.py#L9) is used.

BSchilperoort:
Closes #113 

Passing labels:

- did not work
- was not clear to me what the use-case is (extra labels don't really hurt...)
- was fragile (users need to know what the possible labels are, can't enter a free form prompt
- didn't see how to clearly fix it

So I removed the (broken) functionality

### DIFF
--- a/docs/tutorial/cli.md
+++ b/docs/tutorial/cli.md
@@ -102,7 +102,7 @@ Download Mapillary images to a local directory.
 │ --token               Mapillary OAuth token (if not set via                  │
 │                       MAPILLARY_TOKEN).                                      │
 │ --project             An optional project to attach to.                      │
-╰────────────────────────────────────────────────────────────────────────────────╯
+╰──────────────────────────────────────────────────────────────────────────────╯
 ```
 
 ## Segmenting images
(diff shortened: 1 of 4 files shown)


docs/tutorial/cli.md
```diff
@@ -160,15 +159,15 @@ Segment images with MaskFormer.
...
 │ --project                The project to use. Uses the active project by      │
 │                          default. [default: local-test]                      │
 │ --overwrite              Overwrite an existing run. [default: False]         │
 │ --verbose                Print verbose log to the terminal. Useful for       │
 │                          debugging models. [default: False]                  │
-╰────────────────────────────────────────────────────────────────────────────────╯
+╰──────────────────────────────────────────────────────────────────────────────╯
 ```
 
 By default, all images in the current streetscapes project will be segmented. These will be processed in batches whose size can be specified with the `batch_size` option (the default is `10`) depending on your hardware it can be better to make that number smaller (laptop) or larger (HPC with GPU). You can also specify a (comma-separated) list of labels (categories of objects) that the model should focus on. By default, if the `--labels` argument is not provided, the model will try to find objects corresponding to ***all*** the categories that it can recognise.
```
cantordust:
```suggestion
By default, all images in the current streetscapes project will be segmented. These will be processed in batches whose size can be specified with the `batch_size` option (the default is `10`) depending on your hardware it can be better to make that number smaller (laptop) or larger (HPC with GPU).
```

cantordust (approved):
Makes sense, there are not that many categories anyway.

docs/tutorial/cli.md
BSchilperoort: Good catch, missed that

PR #193, opened 2026-09-04
Store instance confidence

### LINKED ISSUE #184: Store instance confidence if model provides this
Some models (e.g. SAM3) also give a confidence score to the resulting instances.

Example use case:
- Segment w/ SAM3 with the prompt "brick facade, concrete facade, glass facade"
- A building might get both the "brick facade" and "concrete facade" label
- Model confidence can be used to pick between the most likely material

BSchilperoort:
Confidence/model score is now stored for instances when using;

- SAM3
- MaskFormer
- DinoSAM

BFMS does not provide scores

### DIFF
--- a/src/streetscapes/models/dinosam/model.py
+++ b/src/streetscapes/models/dinosam/model.py
@@ -135,6 +135,7 @@ class DinoSAM:
                 # No objects found, but still record the image as processed.
                 logger.debug(f"No objects found in image '{uid}'.")
                 segmentation["labels"] = []
+                segmentation["confidences"] = []
                 segmentation["instances"] = np.zeros(
                     (0, *image.shape[:2]), dtype=np.bool_
                 )
@@ -168,6 +169,9 @@ class DinoSAM:
             # Instance labels from GroundingDINO
             instance_labels = dino_results["text_labels"]
 
+            # Detection confidence reported by GroundingDINO for each instance.
+            confidences = dino_results["scores"].cpu().numpy().astype(float).tolist()
+
             # Populate the instance masks.
             instances = np.zeros(
                 (len(instance_labels), *image.shape[:2]),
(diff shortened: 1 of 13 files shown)


src/streetscapes/project.py
```diff
@@ -88,6 +88,7 @@ class Project:
                 "labels": "STRING[]",
                 "rating": "INTEGER",  # 0-5
                 "polygons": "GEOMETRY",
+                "confidences": "FLOAT8[]",
```
cantordust, line 91: `FLOAT4[]` might be enough here to save some space, I don't think any of the models would output values outside that range.

src/streetscapes/models/dinosam/model.py
```diff
@@ -168,6 +169,9 @@ def segment_images(
             # Instance labels from GroundingDINO
             instance_labels = dino_results["text_labels"]
 
+            # Detection confidence reported by GroundingDINO for each instance.
+            confidences = dino_results["scores"].cpu().numpy().astype(float).tolist()
+
```
cantordust, line 174: Just pointing out that this is the object detection confidence (= how confident the model is that there is an object there). If we want the segmentation confidence (= how confident the model is that a pixel belongs to a certain class), these values should be taken from the SAM model instead.

cantordust (approved):
This is very nice to have. We could add sliders that adjust the masks based on the confidence threshold for different classes, which can be used as part of a fine-tuning pipeline  down the track.

src/streetscapes/project.py
BSchilperoort: Seeing as the polygon geometry contains many many more numbers which are all doubles, I don't think we need to worry about this :sweat_smile:

src/streetscapes/models/dinosam/model.py
BSchilperoort: I think in streetscapes the object detection confidence is more important that the segmentation confidence. If users want to get, e.g., all doors and windows of a building, the confidence of the model that it is in fact a window is the most important (rather than which exact pixels belong to the window).

BSchilperoort:
Thanks for the review!

PR #195, opened 2026-09-04
Split up utils functions module, move to tz aware timestamps

### LINKED ISSUE #157: Break up the `functions` module into submodules
All functions that have some more or less generic utility live under the `functions` module. This was fine while we had 3-4 functions, but now having everything but the kitchen sink makes it somewhat unwieldy.

Functions should be grouped into functional categories into separate submodules.

### LINKED ISSUE #167: Store `captured_at` column as timestamp
For the images, we store the timestamp of the photographs under the `captured_at` column, which currently has a `UBIGINT` format (ms since epoch). However, we can store this in duckdb as a `TIMESTAMP` instead. This would be easier for end users and require less conversion in other places in the code.

see https://github.com/Urban-M4/streetscapes/pull/163/changes#r3289205014

BSchilperoort:
Closes #157 (900a5ed)

- Removed unused is_notebook, show_image functions.
- Duplicate get_env definition was removed.

Also closes #167 (c1ddab9)

- moved to timezone-aware timestamps in database.
- infer timezone from EXIF's gps coordinates

Wanted to break the old databases again as #193 broke them anyway.


cantordust (approved):
Thank you, nice refactoring - I ran a full segmentation pipeline from scratch (fetch, download, segment...), and it all seemed to work.🖖

PR #197, opened 2026-09-09
Add Mapillary API result validation

### LINKED ISSUE #181: Sanitize mapillary metadata API output

BSchilperoort:
Closes #181


BSchilperoort:
Tested by fetching metadata + downloading 3500 images of Amsterdam.

src/streetscapes/sources/mapillary.py
```diff
@@ -22,11 +30,92 @@
 logger = logging.getLogger(__name__)
 
 
+class MapillaryImage(BaseModel):
```
cantordust: Good idea to use a Pydantic model!

src/streetscapes/utils/db_types.py
```diff
@@ -0,0 +1,81 @@
...
+
+    if not (-180.0 <= lon <= 180.0 and -90.0 <= lat <= 90.0):
+        raise ValueError(f"Point coordinates out of range: {coords!r}")
+
+    # WKT facilitates conversion to either GeoPandas or DuckDB geometry.
+    return Point(lon, lat).wkt
+
+
+def epoch_ms_to_datetime(value: Any) -> Any:
+    """Convert milliseconds since the (UTC) epoch into an aware datetime."""
```
cantordust, line 40: Should this be 'timezone-aware'? If yes, should the timezone be passed as an argument?

cantordust (approved):
Good call to use a Pydantic model for validating data coming from Mapillary!

src/streetscapes/utils/db_types.py
BSchilperoort: miliseconds since epoch _should_ be in UTC.

That doesn't mean that the timestamps in Mapillary are accurate. KartaView timestamps are often not.
It depends on what they used to push the images to mapillary and if their camera time was accurate.

BSchilperoort: I've added a note to kartaview docs to inform users; e69a07a

PR #198, opened 2026-09-09
Drop unused segmentation-level notes

### LINKED ISSUE #153: explrore: segmentation notes drop or implement fully
At https://github.com/Urban-M4/streetscapes/blob/950021dcff2646444870151a0dbedc7b97da7a94/src/streetscapes/explorer/data.py#L51

but the db does not contain notes column in segmentations table. Only images table has notes column.

We should either
1. Drop notes from Segmentation dataclass
2. Add notes column to segmentations table + retrieve notes in _get_segmentations function + add post endpoint to update notes

BSchilperoort:
Closes #153


cantordust (approved):
No notes 🤭

PR #201, opened 2026-09-10
Reimplement KartaView fetch+download

BSchilperoort:
LLM did quite OK as structure was already well defined.

Note: first review #197 

**Edit:** KartaView API seems to be quite unstable. #206 adds the option to add a token but this does not fix the issue. It does help with rate limits however.

### DIFF
--- a/docs/tutorial/cli.md
+++ b/docs/tutorial/cli.md
@@ -54,10 +54,12 @@ The `project_dir` directory is where Streetscapes will be storing its projects, 
 
 ## Downloading images
 
-The Streetscapes CLI supports downloading images from [Mapillary](https://www.mapillary.com/) ([KartaView](https://kartaview.org/landing) and the [Amsterdam](https://api.data.amsterdam.nl/) collection are currently not yet supported). The available options can be displayed with the `--help` option via the subcommand for Mapillary:
+The Streetscapes CLI supports downloading images from [Mapillary](https://www.mapillary.com/) and [KartaView](https://kartaview.org/landing) (the [Amsterdam](https://api.data.amsterdam.nl/) collection is currently not yet supported). Both follow the same two steps — fetch the metadata for a bounding box, then download the images it describes. The available options can be displayed with the `--help` option via the subcommand for each source:
+
+### Mapillary
 
 For Mapillary, we first need to fetch image metadata. For this you will need to define a spatial bounding box.
-Some areas have enormous amounts of images available. To only get a certain number of images per spatial "tile", set the `--limit` argument.
+Some areas have enormous amounts of images available. To only get a certain number of images per spatial "tile", set the `--tile-limit` argument.
 
 Note that a [token](https://www.mapillary.com/developer/api-documentation/) is needed to use the Mapillary API.
 Register on Mapillary, and register your token with `streetscapes config set mapillary_token YOUR_TOKEN`.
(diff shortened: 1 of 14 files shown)


BSchilperoort:
Note to self: figure out of shotDate from KartaView API is UTC or local time. KartaView isn't clear about this. 
Retrieving some images from Asia and Central America should help with figuring this out.

BSchilperoort:
> figure out of shotDate from KartaView API is UTC or local time. KartaView isn't clear about this.
> Retrieving some images from Asia and Central America should help with figuring this out.

Image from Wellington NZ:

image

Sun is in the north/north west. Assuming KartaView time is UTC, local time (GMT+12) would be ~2pm. That matches the location of the sun in the sky so I think it's safe to assume shotDate is in UTC.

BSchilperoort:
> Sun is in the north/north west. Assuming KartaView time is UTC, local time (GMT+12) would be ~2pm. That matches the location of the sun in the sky so I think it's safe to assume shotDate is in UTC.

Had llm look into this more. Turns out it's a mess and kartaview is probably not consistent. UTC is best assumption though

<hr>

There's no single answer: shotDate has no fixed timezone. KartaView's docs don't say what it is, and it looks like each uploading app records the time in its own way.

Evidence. Street imagery is nearly always taken in daylight. So for each sequence I checked whether its first image falls in daylight when shotDate is read as UTC, and when it's read as local time:


Region	Sequences	In daylight if UTC	In daylight if local time
Seattle	142	134	24
Wellington	52	37	15
Auckland	38	13	36

Seattle is clearly UTC and Auckland is clearly local time. Wellington is mostly UTC, but still has a sizeable group that only fits local time. The regions with few sequences are probably dominated by one or two uploaders, so this varies by uploader or app, not by place.

docs/tutorial/cli.md
```diff
@@ -54,10 +54,12 @@ The `project_dir` directory is where Streetscapes will be storing its projects,
 
 ## Downloading images
 
-The Streetscapes CLI supports downloading images from [Mapillary](https://www.mapillary.com/) ([KartaView](https://kartaview.org/landing) and the [Amsterdam](https://api.data.amsterdam.nl/) collection are currently not yet supported). The available options can be displayed with the `--help` option via the subcommand for Mapillary:
+The Streetscapes CLI supports downloading images from [Mapillary](https://www.mapillary.com/) and [KartaView](https://kartaview.org/landing) (the [Amsterdam](https://api.data.amsterdam.nl/) collection is currently not yet supported). Both follow the same two steps — fetch the metadata for a bounding box, then download the images it describes. The available options can be displayed with the `--help` option via the subcommand for each source:
```
cantordust, line 57:
```suggestion
The Streetscapes CLI supports downloading images from [Mapillary](https://www.mapillary.com/) and [KartaView](https://kartaview.org/landing) (the [Amsterdam](https://api.data.amsterdam.nl/) collection is currently not yet supported). Both follow the same two steps — fetch the metadata for a bounding box, then download the images it describes. The available options can be displayed with the `--help` option via the subcommand for each source.
```

docs/tutorial/cli.md
```diff
@@ -72,13 +74,13 @@ Usage: streetscapes fetch-metadata mapillary [OPTIONS] BBOX
...
-│ *  BBOX  Bounding box (WEST EAST SOUTH NORTH). [required]                      │
+│ *  BBOX  Bounding box (WEST SOUTH EAST NORTH). [required]                      │
 ╰────────────────────────────────────────────────────────────────────────────────╯
 ╭─ Parameters ───────────────────────────────────────────────────────────────────╮
-│ --tile-size  Tile size in degrees. [default: 0.001]                            │
-│ --limit      Maximum number of images per tile. [default: 1000]                │
-│ --token      Mapillary OAuth token (if not set via MAPILLARY_TOKEN).           │
-│ --project    An optional project to attach to.                                 │
+│ --tile-size   Tile size in degrees. [default: 0.001]                           │
+│ --tile-limit  Maximum number of images per tile. [default: 1000]               │
```
cantordust, line 81: `--tile-limit` sounds like a limit on the number of tiles, not the number of images. Would it be OK to use something like `--images-per-tile`? Although we probably don't want very long switch names...the KartaView subcommand below has a `--image-limit` switch.

docs/tutorial/cli.md
```diff
@@ -105,6 +107,42 @@ Download Mapillary images to a local directory.
...
+The images are then downloaded in the same way:
+
+```bash
+streetscapes download-images kartaview
+```
+
+KartaView serves its images at full resolution, which for recent cameras means 4K.
+Segmenting those needs a correspondingly large amount of memory, because the models
+scale their masks back up to the size of the image they were given — so on a machine
+with limited RAM, segment KartaView images in small batches (`--batch-size 1`).
```
cantordust, line 144: Just a note: this is good advice, but I wonder if we should add an option to resize large images to a maximal size when they are downloaded.

cantordust (approved):
Great, it seems to work as intended!🚀

docs/tutorial/cli.md
BSchilperoort: I think there is a value to having high-res images. Also makes everything related to UUIDs difficult as the resized image has a different UUID...

docs/tutorial/cli.md
BSchilperoort: Yeah it's not optimal. But `--image-limit` is a total image limit.

`--images-per-tile` might be the best option here...

docs/tutorial/cli.md
cantordust: Good point about the UUIDs!

PR #202, opened 2026-09-10
Add myself to CITATION.cff

BSchilperoort:
Thanks for noticing, @sverhoeven 

I think I contributed sufficiently to be added to the CITATION.cff file :sweat_smile:


sverhoeven (approved):
Looks good.

PS You do not need to add email for Zenodo integration, but as your email is already in all the commits it does not matter that much.

PR #203, opened 2026-09-11
Add panoramax, fix tiling bug

### LINKED ISSUE #200: Support Panoramax as image source
[Panoramax](https://panoramax.fr/) is a federation of street-level image initiatives, with a common pictures catalog.

It would be good to support this vs. Mapillary (owned by Meta) and KartaView (depends on [Grab Holdings](https://en.wikipedia.org/wiki/Grab_Holdings)).

BSchilperoort:
Closes #200 

Panoramax is a federated street level images database. It largely contains images of France but also has instances focused on other areas.

For usage see docs/cli.

<hr>

During testing a bug in the tiling logic came up; if you have a bbox which is smaller than a tile, the images returned by the metadata fetcher would still be filtered for your original bbox again, making the per tile image limit not work properly.

This is fixed by trimming the tiles to the bbox extents

### DIFF
--- a/docs/tutorial/cli.md
+++ b/docs/tutorial/cli.md
@@ -54,7 +54,7 @@ The `project_dir` directory is where Streetscapes will be storing its projects, 
 
 ## Downloading images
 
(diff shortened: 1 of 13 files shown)


BSchilperoort:
Panoramax seems to be very fast, and has a nice API with good filtering options. Assuming more images will be added outside of France this seems like the most important source to support long-term (connected to OSM community, federated architecture).

cantordust (approved):
I like Panoramax as a source, thank you for adding support for it!🚀

PR #205, opened 2026-09-11
Add pano-only option to fetch commands

BSchilperoort:
For facade analysis panoramic images might be more useful, as they show facades at both sides of the road. Normal images can mostly show the street surface and vehicles, while buildings are on the edges of the image.

This PR allows users to filter for pano-only images when fetching img metadata. This will reduce the images to (wastefully) download and segment.


cantordust (approved):
Also works as expected, good feature to have.🖖

PR #206, opened 2026-09-14
Daylight filter

BSchilperoort:
- added `--daytime-only` option for metadata fetches
- added optional kartaview token for increased rate limit

Also added some improvements from review of earlier prs


cantordust (approved):
Another filter that's nice to have!🖖

PR #207, opened 2026-09-15
Lower memory SAM3 predictor

BSchilperoort:
SAM3 memory use is quite high. Would be nice to reduce this. 

This PR reduces memory use by fixing the Predictor class, processing 8 masks at a time on GPU instead of all at one. For large image sizes with many masks this greatly reduces RAM usage and can even prevent OOM issues.


BSchilperoort:
Covered by merging this improvement upstream. Ultralytics should be pinned to >=8.4.154

PR #208, opened 2026-09-15
Mask2poly on worker, transfer images as bytes

BSchilperoort:
Performance improvements to reduce Ray memory consumption and host -> worker transfer overhead


cantordust (approved):
This works, although I didn't see any material reduction in memory consumption.

BSchilperoort:
> although I didn't see any material reduction in memory consumption.

Big masks will exist as pickle and 3D numpy array at the same time. This should show as a short peak in the memory consumption. I think you actually see this if you compare the two flame graphs you sent me yesterday.

W/ Ray:
newplot(1)

Direct comms:
newplot

<hr>

These changes will also help to reduce communication overhead.

